Issue #7, opened 2015-09-03
Add option to add hostname


JohnLZeller:
The agent currently has the option to allow adding a hostname in`datadog.conf`. Let's add this option to the configuration screen for the plugin as well.

JohnLZeller:
+1 for this

kbaltrinic:
Actually this seems to be presenting as a bug.  We have both the old agent and the new plugin running side-by-side for evaluation purposes and we are getting events like `soapui-tests succeeded on ci20-jen-001` from the agent and `soapui-tests build #112 succeeded on null` from the plugin.

We love that the new plugin provides some additional tags, but without mapping the event to a host, and hence to all of the host's tags, its actually a big step backwards.  Good work so far though.  These are just  details to be ironed out.

JohnLZeller:
Thanks @kbaltrinic! I've updated the issue to a bug. We are aware of the host name issue presenting as `null` and will be doing another release of the plugin to fix the issue. And thanks for your other feedback as well! :) I'm glad you're liking the plugin. Looking forward to improving this plugin

JohnLZeller:
PR #15 should handle this, and it has been merged and will be included in the 0.2.0 release (the next release)

PR #15, opened 2015-09-22
Improve reliability of determining hostname

JohnLZeller:
As mentioned in #7, the hostname is not being pulled from the Jenkins host in a reliable fashion, and will often present as `null`. This update provides the ability to set a hostname from the configuration panel, and if that is unavailable, the plugin will search for it in the Jenkins hostname environment variable, then the Unix hostname via `/bin/hostname -f`, and finally the localhost hostname, before instead returning null. This also ensures that logging of success/failure in finding a hostname is done.


JohnLZeller:
Tested, and everything looks good

PR #17, opened 2015-09-22
[tags] Add node, remove build_number, and change job_name to job

JohnLZeller:
Having build_number added as a tag for every metric and event is causing too many contexts to be generated. Additionally, it's not even of much use.


JohnLZeller:
Tested, and works well. Node does not show up in tags for started build events, which I suspect is because the function is called before a node is provisioned.

PR #38, opened 2016-03-10
Added ability to add addtional tags. Fixes #11

### LINKED ISSUE #11: Allow adding custom tags per job
The simplest solution would be to add a configuration option to each jobs config page where an admin could enter in a set of tags to tag the job with.  However a static list of tags, while useful, is not sufficient for us.  

We us parameterized jobs extensively. For example the same job builds all of our dropwizard projects in different configuration such as snapshot builds, release, builds, builds targeting platform X, targeting platform Y, etc.  At the very least we would need to be able to specify that certain build parameters, or even certain environment variables, need to be set as the values of certain tags.  For example set the #release tag to the value of the $DO_RELEASE env variable, and the #project tag be set to something based on which git repository we built from.

One possibility that would provide for maximum flexibility would be to be able to read in a set of tag values from a specified file in the workspace.  This way we could use any sort of post build script to programmatically setup our tags.

MadsNielsen:
This is the first proposal for fixing issue #11

The pull request adds the ability to add tags to DataDog.

f72986bc-e5e7-11e5-8a2f-5ca36fd5863d

Tags can be added by reading a file, or entering a properties file that follows the standard java property file syntax: {k}={v}. The value can here be something that is expanded by Jenkins, for example an environment variable or a build parameter or a fixed value.

The tags are added to event, metric and the service check.

Comments are welome.


MadsNielsen:
@JohnLZeller Here you go. In the correct repo. :)

JohnLZeller:
Thanks @MadsNielsen! I'm reviewing this now :)

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/help-tagging.html
```diff
@@ -0,0 +1,6 @@
+<div>
+    <p>A newline separated list of tags to push to DataDog. Tags MUST follow the format <em>key:value</em>.<p>
```
JohnLZeller: Should probably say `key=value`

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/help-tagging.html
```diff
@@ -0,0 +1,6 @@
...
+    <p>Example: <em>os=${OS_PARAM}</em>.</p> 
+    <p>The value can either be something that is automatically expanded in Jenkins, or a static value which you define for the tag</p>
+    <p>The OS_PARAM in this case is a value that is expanded by Jenkins (Environment and/or build variable)</p>
```
JohnLZeller: These two lines could be condensed to say, 

`<p>The value can either be something that is automatically expanded by Jenkins (environment and/or build variable), or a static value. The OS_PARAM in this case is expanded by Jenkins.</p>`

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/help-tagFile.html
```diff
@@ -0,0 +1,6 @@
+<div>
+    <p>Read a file from workspace as a newline seperated list of tags to push to DataDog. Location is relative to workspace</p>
+    <p>Example: <em>os=${OS_PARAM}</em>.</p> 
+    <p>The value can either be something that is automatically expanded in Jenkins, or a static value which you define for the tag</p>
+    <p>The OS_PARAM in this case is a value that is expanded by Jenkins (Environment and/or build variable)</p>
```
JohnLZeller: ditto here

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/config.jelly
```diff
@@ -0,0 +1,14 @@
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:optionalBlock name="on" checked="${instance.on}" title="DataDog tagging" inline="true">
```
JohnLZeller: I'd prefer that we put this in a section of its' own with the heading `Datadog Tagging`

Similar to how Slack does it
screen shot 2016-03-11 at 2 05 44 am

JohnLZeller:
Please make sure you remove any tabs on blank lines, and make sure that all the functions have doc comments.

This is looking pretty good. I'd like to test it a bit more before signing off though :) More tomorrow

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -195,13 +247,12 @@ private JSONObject gatherBuildMetadata(final Run run, @Nonnull final TaskListene
     builddata.put("hostname", getHostname(envVars)); // string
     builddata.put("buildurl", envVars.get("BUILD_URL")); // string
     builddata.put("node", envVars.get("NODE_NAME")); // string
-
+
```
JohnLZeller: tab

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -317,15 +381,17 @@ public final void gauge(final String metricName, final JSONObject builddata,
     metric.put("points", points);
     metric.put("type", "gauge");
     metric.put("host", builddata.get("hostname"));
-    metric.put("tags", assembleTags(builddata));
-
+    metric.put("tags", assembleTags(builddata, extraTags));
+
```
JohnLZeller: tab

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -317,15 +381,17 @@ public final void gauge(final String metricName, final JSONObject builddata,
...
     metric.put("host", builddata.get("hostname"));
-    metric.put("tags", assembleTags(builddata));
-
+    metric.put("tags", assembleTags(builddata, extraTags));
+    
     // Place metric as item of series list
     JSONArray series = new JSONArray();
     series.add(metric);
-
+
```
JohnLZeller: tab

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ * Create a job property for use with DataDog plugin.
+ * @param <T>
+ */
+public class DataDogJobProperty<T extends Job<?,?>> extends JobProperty<T> {
+
+  private static final Logger LOGGER =  Logger.getLogger(DatadogBuildListener.class.getName());
+  private static final String DISPLAY_NAME = "DataDog Tagging";
```
JohnLZeller: "DataDog Job Tagging"

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  /**  
+   * This is a list of tagging targets to be submitted with the Build to DataDog.
+   * @return the tagging
+   */
+  public String getTagging() {
+    return tagging;
+  }
+
+  @DataBoundConstructor
+  public DataDogJobProperty( ) { }
```
JohnLZeller: this should be first, above the `getTagging()` function.

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  public void setTagging(final String tagging) {
+    this.tagging = tagging;
+  }
+  
+  /**
+   * This method is called whenever the Job form is saved. We use the 'on' property 
+   * to determine if the controls are selected.
+   * 
+   * @param req - The request
+   * @param form - The submitted form
```
JohnLZeller: `A JSONObject containing the submitted form data from the job configuration`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+   * @return The configured {@link JobProperty}
```
JohnLZeller: `@return a {@link JobProperty} object representing the tagging added to the job`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  private static final Logger LOGGER =  Logger.getLogger(DatadogBuildListener.class.getName());
+  private static final String DISPLAY_NAME = "DataDog Tagging";
+  private String tagging = null;  
+  private String tagFile = null;
+  
+  /**  
+   * This is a list of tagging targets to be submitted with the Build to DataDog.
+   * @return the tagging
+   */
+  public String getTagging() {
```
JohnLZeller:
```
  /**
   * Gets a list of tagging targets to be submitted with the Build to DataDog.
   *
   * @return a String representing a list of tagging targets.
   */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+   * @return the tagFile
+   */
+  public String getTagFile() {
+    return tagFile;
+  }
+
+  /**
+   * @param tagFile the tagFile to set
+   */
+  @DataBoundSetter
```
JohnLZeller:
```

  /**
   * Sets a list of tagging targets to be submitted with the Build to DataDog.
   *
   * @param tagging - A String representing a list of tagging targets.
   */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+        prop.tagging = null;
+    }
+    
+    return prop;
+  }
+
+  /**
+   * @return the tagFile
+   */
+  public String getTagFile() {
```
JohnLZeller:
```

  /**
   * Gets the tagFile set in the job configration.
   *
   * @return a String representing the relative path to a tagFile
   */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+   */
+  public String getTagging() {
+    return tagging;
+  }
+
+  @DataBoundConstructor
+  public DataDogJobProperty( ) { }
+
+  @DataBoundSetter
+  public void setTagging(final String tagging) {
```
JohnLZeller:
```

  /**
   * Sets a list of tagging targets to be submitted with the Build to DataDog.
   *
   * @param tagging - A String representing a list of tagging targets.
   */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * @return A boolean indicating if 'Properties content' was set.
+   */  
+  public boolean isTaggingEmpty() {
+    return StringUtils.isBlank(tagging);
+  }
+  
+  /**
+   * Method to read the contents of the specified file in the {@link DataDogJobProperty}
+   * @param r - Current build
```
JohnLZeller: nit: should be an empty comment line between description and param

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  @DataBoundSetter
+  public void setTagFile(String tagFile) {
+    this.tagFile = tagFile;
+  }
+  
+  /**
+   * @return A boolean indicating if 'File from workspace' was set.
+   */
+  public boolean isTagFileEmpty() {
+    return StringUtils.isBlank(tagFile);
```
JohnLZeller: I think this needs to be `this.tagFile`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+   */
+  @DataBoundSetter
+  public void setTagFile(String tagFile) {
+    this.tagFile = tagFile;
+  }
+  
+  /**
+   * @return A boolean indicating if 'File from workspace' was set.
+   */
+  public boolean isTagFileEmpty() {
```
JohnLZeller:
```

  /**
   * Checks if tagFile was set in the job configuration
   *
   * @return a boolean representing the state of the tagFile job configuration
   */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+    return StringUtils.isBlank(tagFile);
+  }
+
+  /**
+   * @return A boolean indicating if 'Properties content' was set.
+   */  
+  public boolean isTaggingEmpty() {
```
JohnLZeller:
```
 /**
   * Checks if the contents of the properties in the job tagging configuration section is empty
   *
   * @return a boolean representing the state of the properties job configuration
   */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Extension
+  public static final class DataDogJobPropertyDescriptorImpl 
+    extends JobPropertyDescriptor {
+      
+    /**
+     * @return The display name of the JobProperty.
+     */  
+    @Override
```
JohnLZeller:
```
    /**
     * Getter function for a human readable class display name.
     *
     * @return a String containing the human readable display name for the {@link JobProperty} class.
     */
```

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+    public String getDisplayName() {
+        return DISPLAY_NAME;
+    }
+    
+    /**
+     * Indicates where this property can be used
+     * 
+     * @param jobType
```
JohnLZeller: `@param jobType - a Job object

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -85,6 +87,60 @@
...
+   * Always returns a HashMap, that can be empty, if no tagging is configured.
+   * 
+   * @param run - Current build
+   * @param listener - Current listener
+   * @return A {@link HashMap} containing the key,value pairs of tags. Never null.
+   * @throws IOException
+   * @throws InterruptedException 
+   */
+  @Nonnull
+  private HashMap<String,String> parseTagList(Run run, TaskListener listener) throws IOException,
```
JohnLZeller: Let's move this method to below the `assembleTags()` method.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -85,6 +87,60 @@
...
+          InterruptedException {
+    HashMap<String,String> map = new HashMap<String, String>();
+    
+    DataDogJobProperty property = (DataDogJobProperty)run.getParent()
+            .getProperty(DataDogJobProperty.class);
+    String prop = property.getTagging();
+    
+    PrintStream console = listener.getLogger();
```
JohnLZeller: remove this, let's use the `logger` for the class

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -85,6 +87,60 @@
...
+    
+    if( !property.isTagFileEmpty() ) {
+      String dataFromFile = property.readTagFile(run);
+      if(dataFromFile != null) {
+        for(String tag : dataFromFile.split("\\r?\\n")) {    
+          String[] expanded = run.getEnvironment(listener).expand(tag).split("=");
+          if( expanded.length > 1 ) {
+            map.put(expanded[0], expanded[1]);
+            console.println(String.format("Emitted tag %s:%s", expanded[0], expanded[1]));
```
JohnLZeller: Use `logger.fine(...)`

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -85,6 +87,60 @@
...
+        }
+      }
+    }
+      
+    if( !property.isTaggingEmpty() ) {
+      for(String tag : prop.split("\\r?\\n")) {    
+        String[] expanded = run.getEnvironment(listener).expand(tag).split("=");
+        if( expanded.length > 1 ) {
+          map.put(expanded[0], expanded[1]);
+          console.println(String.format("Emitted tag %s:%s", expanded[0], expanded[1]));
```
JohnLZeller: Use `logger.fine(...)`

JohnLZeller:
Thanks for the updates :) I added a bunch of nitpicks, mostly for good javadoc content.

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/config.jelly
```diff
@@ -0,0 +1,20 @@
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:section title="DataDog Tagging"> 
+        <f:optionalBlock name="enableFile" checked="${!instance.isTagFileEmpty()}" title="File from workspace" inline="true">
```
JohnLZeller: I'm thinking that the options could have better names. Maybe the title of this one could become `Add tags from file in workspace`.

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/config.jelly
```diff
@@ -0,0 +1,20 @@
...
+            <f:entry title="File" field="tagFile">
+                <f:textbox/>
+            </f:entry>
+        </f:optionalBlock>
+        <f:optionalBlock name="enableProperty" checked="${!instance.isTaggingEmpty()}" title="Properties content" inline="true">
```
JohnLZeller: Maybe this title could become `Add tags from list of properties`

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/config.jelly
```diff
@@ -0,0 +1,20 @@
...
+            <f:entry title="Properties" field="tagging">
```
JohnLZeller: This field name isn't self-descriptive enough. Let's go with `tagProperties`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  private static final String DISPLAY_NAME = "DataDog Tagging";
+  private String tagging = null;  
+  private String tagFile = null;
+  
+  /**  
+   * This is a list of tagging targets to be submitted with the Build to DataDog.
+   * @return the tagging
+   */
+  public String getTagging() {
+    return tagging;
```
JohnLZeller: `tagProperties`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Create a job property for use with DataDog plugin.
+ * @param <T>
+ */
+public class DataDogJobProperty<T extends Job<?,?>> extends JobProperty<T> {
+
+  private static final Logger LOGGER =  Logger.getLogger(DatadogBuildListener.class.getName());
+  private static final String DISPLAY_NAME = "DataDog Tagging";
+  private String tagging = null;
```
JohnLZeller: `tagProperties`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+  public String getTagging() {
+    return tagging;
+  }
+
+  @DataBoundConstructor
+  public DataDogJobProperty( ) { }
+
+  @DataBoundSetter
+  public void setTagging(final String tagging) {
+    this.tagging = tagging;
```
JohnLZeller: `this. tagProperties = tagProperties;`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+   * @throws hudson.model.Descriptor.FormException 
+   */
+  @Override
+  public JobProperty<?> reconfigure(StaplerRequest req, JSONObject form) 
+          throws Descriptor.FormException {
+    
+    DataDogJobProperty prop = (DataDogJobProperty) super.reconfigure(req, form);
+    
+    boolean isEnableFile = form.getBoolean("enableFile");
+    boolean isEnableTag = form.getBoolean("enableProperty");
```
JohnLZeller: `isEnableTagProperties`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+    
+    if(!isEnableFile) {
+        prop.tagFile = null;
+    } 
+    if(!isEnableTag) {
```
JohnLZeller: `isEnableTagProperties`

src/main/java/org/datadog/jenkins/plugins/datadog/DataDogJobProperty.java
```diff
@@ -0,0 +1,148 @@
...
+        prop.tagging = null;
```
JohnLZeller: `tagProperties`

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -85,6 +87,60 @@
...
+   * @throws InterruptedException 
+   */
+  @Nonnull
+  private HashMap<String,String> parseTagList(Run run, TaskListener listener) throws IOException,
+          InterruptedException {
+    HashMap<String,String> map = new HashMap<String, String>();
+    
+    DataDogJobProperty property = (DataDogJobProperty)run.getParent()
+            .getProperty(DataDogJobProperty.class);
+    String prop = property.getTagging();
```
JohnLZeller: `String prop = property.getTagProperties();`

JohnLZeller:
Sorry for all the comments @MadsNielsen :) Found some more.

MadsNielsen:
That's fine.You're the PO on this one :). One question though. Is there a reason why you're not pulling tags from the official Jenkins release repo?

JohnLZeller:
We could, but this repo is used for development (including issues and prs), and we just merge to the jenkinsci hosted repo for "official" releases. Hence, why tags show up in one and not the other.

JohnLZeller:
I'll try to get this reviewed today/tomorrow

irabinovitch:
@JohnLZeller  Were you able to take a look at this one?

JohnLZeller:
Yup! Reviewing it. I have a few comments left to give, but this looking
almost g2g.

JohnLZeller:
Since we changed a field name to `tagProperties` the resource `help-tagging.html` needs to be renamed to `help-tagProperties.html`

MadsNielsen:
@JohnLZeller There you go. I fixed the commit up. I found an option to remove trailing white spaces in my editor...seems to have cached them...:)

MadsNielsen:
@JohnLZeller Hold on getting this merged in. I've found a bug i need to investigate.

MadsNielsen:
@JohnLZeller I have fixed the bug i found. 

We need to check for the existence of a a workspace before looking for the file when looking for tags in a file. When onStarted() is invoked in the runListener the workspace is not present. It is later on when metrics etc...are emitted. 

So i added a simple check in readTagFile like so: 

```
  public String readTagFile(Run r) {
    String s = null;
    try {
      //We need to make sure that the workspace has been created. When 'onStarted' is
      //invoked, the workspace has not yet been established, so this check is necessary.
      FilePath workspace = r.getExecutor().getCurrentWorkspace();
      if(workspace != null) {
        FilePath path = new FilePath(workspace,
                tagFile);
        if(path.exists()) {
          s = path.readToString();
        }
      }
    } catch (IOException ex) {
      LOGGER.severe(ex.getMessage());
    } catch (InterruptedException ex) {
      LOGGER.severe(ex.getMessage());
    }
    return s;
  }
```

JohnLZeller:
@MadsNielsen thanks for catching that! It looks like there is one more thing needed. Tag properties are being pulled in properly and the tags show up on both started and completed events. However, tags from the tag file thrown into the workspace are not being included in started events, but only in completed events. This is related to the `Run run` object. It looks like the tag properties are not returned correctly, but `property.isTagFileEmpty()` seems to correctly report that the file is not empty.

MadsNielsen:
@JohnLZeller You're right. And i know why.

During my testing of tagFile (From file) i initially made a version that would look for a file using an absolute path to a local file, this worked fin. That was later changed to use the workspace and the FilePath from Jenkins + a git repo with a properties file.  

The issue is that in the Jenkins API and the RunListener class, ```onStarted(Run r)``` is called _before_ a workspace has been assigned and an SCM checkout has been performed, i realized that when i read your emssage. That is also the reason you see that tags are added properly the the completed events, because when  ```onFinished(Run r)```   is exectued, we already have a checkout and the job _has_ a workspace.

I'm looking into how to fix this, and i think the correct solution, if we want to keep the current setup, where there is no interaction required for the user to monitor all jobs on Jenkins (since we're using a listener) i see two options:

**Option 1**

We move the posting of the _build start_ event to the ```onCheckout()``` (invoked right after checkout, but before _actual_ build starts) method of the SCMListener class. The SCMListener is registered just like the RunListener.

The obvious consequence of this is that there will be a bit of lag added as to when the event is reported to Datadog (the time between when the the job is in the queue on Jenkins and the checkout is finished) but the starttime (```date_happened```) on the event will still be recorded correctly since the start date is stored on the run object that is passed along.  

**Option 2** 

There is also the possibility to wait a bit, and in the beginning only support the use of absolute paths, for example using UNC style file paths to look on a remote file system or simply just document how the current behaviour is. 

Which one do you prefer, obviously Option 1 is the most fun one to implment, but it has some consequences to the current behaviour.

JohnLZeller:
@MadsNielsen thanks for the breakdown! I think Option 1 sounds promising, and correct me if I am wrong, but the time at which checkout happens is when it leaves the queue, and the job begins in earnest. If this is the case, then I think it is reasonable that we use this object instead, as long as (as you say) the date_happened remains consistent with the actual start of the job.

acasademont:
Hi all! Would this allow us to insert a tag for the Project the Job is in? We have Project X and a bunch of jobs in it. Right now we can't group those jobs per project. I guess that would be something like `final String projectName =  run.getParent().getParent().getDisplayName()` directly in Java. Maybe should I open a separate PR for that?

JohnLZeller:
@acasademont yeah open a separate PR for that. This PR would allow you to specify a custom tag based on things like string literals and jenkins env vars. Seems like project could be a use case for an optional tag. Thanks for your question! :)

MadsNielsen:
@JohnLZeller Yes, when checkout starts the build has left the queue and has been assigned an executor. Would you be ok with this, even when the checkout takes a very long time? What if checkout takes half an hour?

Checkout establishes the jobs workspace which we will need if we want to read the tag values and associate them even with the build-started event in datadog, but since checkout happens later than the onBuildStarted handler in jenksins, it's a chicken and egg problem :).

Would another approach be possible where we add another event, 'Checkout finished' event with all the tag values associated instead of trying to 'cheat' by delaying the reporting to datadog?

JohnLZeller:
@MadsNielsen it actually seems (based on some testing I did) like the started event does not actually get sent until after the job leaves the build queue. The test I just did had a job hanging out in the build queue for 5 minutes, with no job started event, and right after it was given an executor and the build began, the started event popped up.

Given this, it seems like there is very little delay between the two. Let's go for it, and use `onCheckout()`, since we want to be able to associate the events by these tags. Let's also add a note in the use of `onCheckout()` briefly explaining why we are not using `onStarted()`.

MadsNielsen:
@JohnLZeller Super, if you're OK with doing it onCheckout() as in ffter the chosen's job workspace has been established and files from the the SCM is in the jobs worskpace and the delay it can occur i'll go ahead and implement that change. :)  most people measure the SCM operation in seconds.

JohnLZeller:
@MadsNielsen okay let's do this. If the user supplies a set of tagging properties directly, then let's use `onStarted()`. If they supply a tagging file in the workspace, then let's use `onCheckout()`, but have a fall back built-in, in the event that checkout fails. How does that sound?

MadsNielsen:
@JohnLZeller That sounds like a reasonable compromise, we can document the behaviour in the help files of the plugin. I'll go ahead and implement it for you.

JohnLZeller:
@MadsNielsen great! Thanks :)

MadsNielsen:
Hey  @JohnLZeller 

Storry to talk to you again,hope you're ok with me querying all the time :), i've hit yet another snag...it actually relates to the 'fallback' you mentioned, what do we do in the case a checkout fails and the onCheckout() doesn't get invoked? 

As it is...there is simply put nothing we can do, using a listener, unless i implement a new method in Jenkins core (onPostCheckout() on the SCMListener class) which would mean that i would have to bump core requirements to a future Jenkins and we'd have to wait for the PR to be merged.   

I brought up the issue on jenkins-ci dev mailing list, and i asked the question, in the jenkins model, if you want to use the 'configuration-less' monitoring we're using, aside from the Global configuration for the listener, there is no way to fall back in the proces so to speak and report on something that happened in the past. 

So the fallback solution we talked about won't work, 

That leaves me the question: What use cases would be blocked by the tags from the file not showing up on the 'Build start' event as that is what is blocking this feature? Wouldn't that be 'Good enough' for a first version and just document the behaviour in the help files, the way i understood is that this feature is mostly used when metrics are emitted which happens when we do have a workspace?

JohnLZeller:
@MadsNielsen is there something available along the lines of `onCheckoutFailed()`? This would be a good spot to throw this failover event, imo.

Absent that, then I think the best option would be to add an optional configuration.

Using `onStarted()` when a user supplies tagging properties directly.
And when a user supplies the tagging file in the workspace, using `onStarted()`, and adding a checkbox under the filename config on the build configuration page, that allows the user to enable `onCheckout()` events as well. Also adding a help box that explains why this option exists (missing tags in the `onStarted()` events), and then a warning that these events *could* fail to fire in the event of a failed checkout.

There would also need to be a visible warning (and a note in the help box) for the tagging filename, which warns that use of this feature means that start events will not have these tags, but that completed events will.

Let me know if the checkout failed option is viable, and if not, then let's "failover" ( :P ) to this second option.

JohnLZeller:
@MadsNielsen is there any more information that you need from for us to get this out the door?

MadsNielsen:
@JohnLZeller No i think you've done an excellent explantion of what needs to be done. 

Speaking of it. I started on this feature, and in order to share functionality between i had to extract all the methods that i had to reuse between the BuildListener and the SCMListener. 

Are you ok with me doing the following:

Extracting all communications methods / utility methods to a a new class, and make it static to begin with. (methods such as post()...etc, nullSafeString() etc)?

Second note: 
```
  public final void event(final JSONObject builddata) {
    logger.fine("Sending event");

    // Gather data
    JSONObject payload = new JSONObject();
    String hostname = nullSafeGetString(builddata, "hostname");
    String number = nullSafeGetString(builddata, "number");
    String buildurl = nullSafeGetString(builddata, "buildurl");
    String job = nullSafeGetString(builddata, "job");
    long timestamp = builddata.getLong("timestamp");
    String message = "";

    // Setting source_type_name here, to allow modification based on type of event
    payload.put("source_type_name", "jenkins");

    // Build title
    StringBuilder title = new StringBuilder();
    title.append(job).append(" build #").append(number);
    if ( "SUCCESS".equals( builddata.get("result") ) ) {
      title.append(" succeeded");
      payload.put("alert_type", "success");
      message = "%%% \n [See results for build #" + number + "](" + buildurl + ") ";
    } else if ( builddata.get("result") != null ) {
      title.append(" failed");
      payload.put("alert_type", "failure");
      message = "%%% \n [See results for build #" + number + "](" + buildurl + ") ";
    } else {
      title.append(" started");
      payload.put("alert_type", "info");
      message = "%%% \n [Follow build #" + number + " progress](" + buildurl + ") ";
    }
    title.append(" on ").append(hostname);

    // Add duration
    if ( builddata.get("duration") != null ) {
      message = message + durationToString(builddata.getDouble("duration"));
    } 
``` 
Theres logic mixed here with presentation, i'd need to craft a new event that uses the words 'Checkout finished'...would you be ok with me refactoring it so that each event is an object itself and the 'message' you'd need it to append to the package would be a method on the object instead of passing down a new variable to this method, increasing it's complexity?

JohnLZeller:
@MadsNielsen I'm okay with extracting all communications methods / utility methods to a new class. And I think your second point works as well, iiuc. It sounds like you just want a simpler event structure here, that is more flexible, and I think that's worth refactoring for.

MadsNielsen:
@JohnLZeller I've added a second commit to this. That tries to make it more 'Object' oriented. By introducing objects for all the events we want to end to Datadog.

It also adds the stuff we discussed about the optional checkbox to emit an event on after checkout, to give our plugin a chance to read the files in the workspace. 

Can you give it a go and try it out?

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/config.jelly
```diff
@@ -0,0 +1,24 @@
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:section title="DataDog Tagging">
+        <f:optionalBlock name="enableFile" checked="${!instance.isTagFileEmpty()}" title="Add tags from file in workspace" inline="true">
+            <f:entry title="File" field="tagFile">
+                <f:textbox/>
+            </f:entry>
+            <f:entry title="Emit an extra event after checkout" field="emitOnCheckout"
+description="Note: Check this if you want to emit an extra event after checkout. Files are not immediately available when the build leaves the queue. If a checkout fails, the event will not be emitted." >
```
JohnLZeller: Let's modify this title and description, to these:

Title: `Send extra event after a successful checkout`
Description: `Note: Tags set this way, in the workspace, won't appear in started events. To provide visibility of a jobs lifespan, this option will send an extra event, in addition to the started and completed events.`

src/main/resources/org/datadog/jenkins/plugins/datadog/DataDogJobProperty/help-tagFile.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    <p>Read a file from workspace as a newline seperated list of tags to push to DataDog. Location is relative to workspace.</p>
+    <p>Example: <em>os=${OS_PARAM}</em>.</p> 
+    <p>The value can either be something that is automatically expanded by Jenkins (environment and/or build variable), or a static value. The OS_PARAM in this case is expanded by Jenkins.</p>
```
JohnLZeller: Let's add a note about the implications of absent tags as well.

`Note: Tags set this way, in the workspace, won't appear in started events. To provide visibility of a jobs lifespan, check the option for sending an extra event after each successful checkout.`

JohnLZeller:
@MadsNielsen This is looking really good!

I may have mentioned it before, but we should make all the file names and class names which have `DataDog` in them, changed to `Datadog` (without the camel case). Seems like a nitpick, but it's the pattern we follow elsewhere.

Other than that, I noticed an error that showed up when I first installed the plugin, and configured it. It didn't show up again after that. It's possible that it was something that showed up before, which we missed. Could you make sure it's not coming from your changes?

```
Apr 27, 2016 6:25:14 PM hudson.model.UpdateCenter$DownloadJob run
INFO: Installation successful: datadog
Apr 27, 2016 6:25:41 PM hudson.ExpressionFactory2$JexlExpression evaluate
WARNING: Caught exception evaluating: h.getPasswordValue(attrs.value ?: instance[attrs.field]) in /configure. Reason: java.lang.NullPointerException
java.lang.NullPointerException
	at org.datadog.jenkins.plugins.datadog.DatadogBuildListener$DescriptorImpl.getApiKey(DatadogBuildListener.java:459)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.commons.jexl.util.PropertyExecutor.execute(PropertyExecutor.java:125)
	at org.apache.commons.jexl.util.introspection.UberspectImpl$VelGetterImpl.invoke(UberspectImpl.java:314)
	at org.apache.commons.jexl.parser.ASTArrayAccess.evaluateExpr(ASTArrayAccess.java:185)
	at org.apache.commons.jexl.parser.ASTArrayAccess.value(ASTArrayAccess.java:118)
	at org.apache.commons.jexl.parser.ASTReference.execute(ASTReference.java:77)
	at org.apache.commons.jexl.parser.ASTReference.value(ASTReference.java:57)
	at org.apache.commons.jexl.parser.ASTExpression.value(ASTExpression.java:54)
	at org.apache.commons.jexl.parser.ASTElvisNode.value(ASTElvisNode.java:40)
	at org.apache.commons.jexl.parser.ASTExpression.value(ASTExpression.java:54)
	at org.apache.commons.jexl.parser.ASTMethod.execute(ASTMethod.java:81)
	at org.apache.commons.jexl.parser.ASTReference.execute(ASTReference.java:83)
	at org.apache.commons.jexl.parser.ASTReference.value(ASTReference.java:57)
	at org.apache.commons.jexl.parser.ASTReferenceExpression.value(ASTReferenceExpression.java:51)
	at org.apache.commons.jexl.ExpressionImpl.evaluate(ExpressionImpl.java:80)
	at hudson.ExpressionFactory2$JexlExpression.evaluate(ExpressionFactory2.java:74)
	at org.apache.commons.jelly.expression.ExpressionSupport.evaluateRecurse(ExpressionSupport.java:61)
	at org.apache.commons.jelly.expression.ExpressionSupport.evaluateAsString(ExpressionSupport.java:46)
	at hudson.util.jelly.MorphTagLibrary$1.run(MorphTagLibrary.java:109)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.TagSupport.invokeBody(TagSupport.java:161)
	at org.apache.commons.jelly.tags.core.WhenTag.doTag(WhenTag.java:46)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.TagSupport.invokeBody(TagSupport.java:161)
	at org.apache.commons.jelly.tags.core.ChooseTag.doTag(ChooseTag.java:38)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.JellyViewScript.run(JellyViewScript.java:95)
	at org.kohsuke.stapler.jelly.IncludeTag.doTag(IncludeTag.java:147)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.TagSupport.invokeBody(TagSupport.java:161)
	at org.apache.commons.jelly.tags.core.OtherwiseTag.doTag(OtherwiseTag.java:41)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.TagSupport.invokeBody(TagSupport.java:161)
	at org.apache.commons.jelly.tags.core.ChooseTag.doTag(ChooseTag.java:38)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.TagSupport.invokeBody(TagSupport.java:161)
	at org.apache.commons.jelly.tags.core.ForEachTag.doTag(ForEachTag.java:150)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$1.run(CoreTagLibrary.java:98)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.CallTagLibScript$1.run(CallTagLibScript.java:99)
	at org.apache.commons.jelly.tags.define.InvokeBodyTag.doTag(InvokeBodyTag.java:91)
	at org.apache.commons.jelly.impl.TagScript.run(TagScript.java:269)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.kohsuke.stapler.jelly.ReallyStaticTagLibrary$1.run(ReallyStaticTagLibrary.java:99)
	at org.apache.commons.jelly.impl.ScriptBlock.run(ScriptBlock.java:95)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.CallTagLibScript.run(CallTagLibScript.java:120)
	at org.apache.commons.jelly.tags.core.CoreTagLibrary$2.run(CoreTagLibrary.java:105)
	at org.kohsuke.stapler.jelly.JellyViewScript.run(JellyViewScript.java:95)
	at org.kohsuke.stapler.jelly.DefaultScriptInvoker.invokeScript(DefaultScriptInvoker.java:63)
	at org.kohsuke.stapler.jelly.DefaultScriptInvoker.invokeScript(DefaultScriptInvoker.java:53)
	at org.kohsuke.stapler.jelly.JellyFacet$1.dispatch(JellyFacet.java:95)
	at org.kohsuke.stapler.Stapler.tryInvoke(Stapler.java:746)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:876)
	at org.kohsuke.stapler.Stapler.invoke(Stapler.java:649)
	at org.kohsuke.stapler.Stapler.service(Stapler.java:238)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:848)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:686)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1494)
	at hudson.util.PluginServletFilter$1.doFilter(PluginServletFilter.java:132)
	at hudson.util.PluginServletFilter.doFilter(PluginServletFilter.java:123)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1482)
	at hudson.security.csrf.CrumbFilter.doFilter(CrumbFilter.java:49)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1482)
	at hudson.security.ChainedServletFilter$1.doFilter(ChainedServletFilter.java:84)
	at hudson.security.ChainedServletFilter.doFilter(ChainedServletFilter.java:76)
	at hudson.security.HudsonFilter.doFilter(HudsonFilter.java:171)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1482)
	at org.kohsuke.stapler.compression.CompressionFilter.doFilter(CompressionFilter.java:49)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1482)
	at hudson.util.CharacterEncodingFilter.doFilter(CharacterEncodingFilter.java:81)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1482)
	at org.kohsuke.stapler.DiagnosticThreadNameFilter.doFilter(DiagnosticThreadNameFilter.java:30)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1474)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:499)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:137)
	at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:533)
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:231)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1086)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:428)
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:193)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1020)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:135)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:116)
	at org.eclipse.jetty.server.Server.handle(Server.java:370)
	at org.eclipse.jetty.server.AbstractHttpConnection.handleRequest(AbstractHttpConnection.java:489)
	at org.eclipse.jetty.server.AbstractHttpConnection.headerComplete(AbstractHttpConnection.java:949)
	at org.eclipse.jetty.server.AbstractHttpConnection$RequestHandler.headerComplete(AbstractHttpConnection.java:1011)
	at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:644)
	at org.eclipse.jetty.http.HttpParser.parseAvailable(HttpParser.java:235)
	at org.eclipse.jetty.server.AsyncHttpConnection.handle(AsyncHttpConnection.java:82)
	at org.eclipse.jetty.io.nio.SelectChannelEndPoint.handle(SelectChannelEndPoint.java:668)
	at org.eclipse.jetty.io.nio.SelectChannelEndPoint$1.run(SelectChannelEndPoint.java:52)
	at winstone.BoundedExecutorService$1.run(BoundedExecutorService.java:77)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

MadsNielsen:
@JohnLZeller I have looked at your error...and i think i know what the issue is. The method `getApiKey()` uses the `getPlainText()` method on the `Secret` object. The javadoc stats that it can throw a null-pointer exception. That...is probably the case the first time you open (which causes it to call the `get` methods on the descriptor) and configure the plugin.

So i changed it to the more defensive `Secret.toString(apiKey)` which produces the same output but does not throw an exception when the Api key is unconfigured. 

I've also committed the changes you suggested including renaming. Be aware that since we change class names of the stuff i added, you might need to reconfigure the Tagging options on your job.

JohnLZeller:
@MadsNielsen everything looks great! :) The *only* thing, is you missed renaming the `DataDogJobProperty` directory (in resources) to `DatadogJobProperty` :)

MadsNielsen:
@JohnLZeller Gotta love windows case insensitivty! i'll fix this on my linux machine tomorrow :)

JohnLZeller:
@MadsNielsen this looks great! Thanks for all your work on it :) Merging and will release with next version bump

PR #44, opened 2016-05-30
Fixed a possible NPE that can occur when properties are not turned on

MadsNielsen:
Hi @JohnLZeller 

I made this pull request because there is a potential for a NPE when DataDog Job Properties are not selected. 

I think this one is safe to merge and include.


JohnLZeller:
@MadsNielsen thanks!

PR #47, opened 2016-07-01
Use Jenkins source type for all events

olivielpeau:
Was implemented in f00b261 but got reverted (not intentionally afaik) when #38 was merged


JohnLZeller:
Thanks @olivielpeau

Issue #54, opened 2016-08-08
Support for Japanese Job Name


kitamurakei:
Jenkins job name of Japanese will raise garbled character in Datadog.

Like this.
![image](https://cloud.githubusercontent.com/assets/5388202/17476108/3cf748d6-5d99-11e6-998c-adf188cdf0be.png)

There should be is this.
![image](https://cloud.githubusercontent.com/assets/5388202/17476135/65d17510-5d99-11e6-8cbd-b351cccdd34c.png)

JohnLZeller:
Resolved with https://github.com/DataDog/jenkins-datadog-plugin/pull/63

PR #55, opened 2016-08-08
Use OutputStreamWriter and simpleJobName pass to isJobTracked method

kitamurakei:
Hi @JohnLZeller

I made this pull request because #54 .


src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -93,7 +93,7 @@ public final void onStarted(final Run run, final TaskListener listener) {
     String jobName = run.getParent().getDisplayName();
```
todoa2c: This variable name `jobName` should be `displayName`.

JohnLZeller:
Thanks @kitamurakei for the contribution!
Rebased to fix merge commits https://github.com/DataDog/jenkins-datadog-plugin/pull/63

PR #56, opened 2016-08-09
Add whitelist

bhavanki:
This commit adds a whitelist to the plugin to accompany the existing blacklist. If no whitelist is defined, then it has no effect. If it is defined, then the plugin only "tracks" jobs that are listed. If a job is listed in both the blacklist and whitelist, the blacklist wins (the job is not tracked).

I made this change because I found that the plugin picks up each module in a multi-module Maven build as a separate job. This caused the plugin to spam Datadog with lots of internal and (for me) unhelpful information. With the whitelist, I can target the top-level (actual) jobs only.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
 
   <groupId>org.datadog.jenkins.plugins</groupId>
   <artifactId>datadog</artifactId>
-  <version>0.5.1-SNAPSHOT</version>
+  <version>0.5.2-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Datadog Plugin</name>
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
 
   <groupId>org.datadog.jenkins.plugins</groupId>
   <artifactId>datadog</artifactId>
-  <version>0.5.2-SNAPSHOT</version>
+  <version>0.5.3-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Datadog Plugin</name>
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
 
   <groupId>org.datadog.jenkins.plugins</groupId>
   <artifactId>datadog</artifactId>
-  <version>0.5.0-SNAPSHOT</version>
+  <version>0.5.1-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Datadog Plugin</name>
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
 
   <groupId>org.datadog.jenkins.plugins</groupId>
   <artifactId>datadog</artifactId>
-  <version>0.5.3-SNAPSHOT</version>
+  <version>0.5.4-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Datadog Plugin</name>
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
 
   <groupId>org.datadog.jenkins.plugins</groupId>
   <artifactId>datadog</artifactId>
-  <version>0.4.1-SNAPSHOT</version>
+  <version>0.5.0-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Datadog Plugin</name>
--- a/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
+++ b/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
@@ -579,9 +579,9 @@ public class DatadogBuildListener extends RunListener<Run>
     }
 
     /**
-     * Getter function for the optional tag {@link node} global configuration.
+     * Getter function for the optional tag {@link tagNode} global configuration.
      *
-     * @return a Boolean containing optional tag value for the {@link node} global configuration.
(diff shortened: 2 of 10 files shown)


JohnLZeller:
Hey @bhavanki, thanks for your contribution! And I apologize for the lengthy delay.

Would you mind rebase your commit?

In addition, I noticed that this seems to cause jobs to hang, if you install the plugin and never use the whitelist option. Could you look into why that is happening.

I also added some more feedback on here

pom.xml
```diff
@@ -13,7 +13,7 @@
 
   <groupId>org.datadog.jenkins.plugins</groupId>
   <artifactId>datadog</artifactId>
-  <version>0.5.3-SNAPSHOT</version>
+  <version>0.5.4-SNAPSHOT</version>
```
JohnLZeller: Let's not bump this, we do this when we are ready to actually bump the version.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -579,6 +586,17 @@ public String getBlacklist() {
...
+     * Getter function for the {@link whitelist} global configuration, containing
+     * a comma-separated list of jobs to whitelist for monitoring. An empty or missing
+     * list means all jobs are whitelisted.
+     *
+     * @return a String array containing the {@link whitelist} global configuration.
+     */
+    public String getWhitelist() {
+      return whitelist;
+    }
+
```
JohnLZeller: Could we also add a `setWhitelist()`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -57,22 +57,32 @@ public static  String getApiKey() {
...
 
   /**
-   * Checks if a jobName is blacklisted, or not.
+   * Checks if a jobName is blacklisted, whitelisted, or neither.
    *
    * @param jobName - A String containing the name of some job.
-   * @return a boolean to signify if the jobName is or is not blacklisted.
+   * @return a boolean to signify if the jobName is or is not blacklisted or whitelisted.
    */
   public static boolean isJobTracked(final String jobName) {
```
JohnLZeller: The functionality of `isJobTracked()` seemed inconsistent with the desired operation, so I played with it. Something like this would work better I think

```python

  /**
   * Checks if a jobName is blacklisted, whitelisted, or neither.
   *
   * @param jobName - A String containing the name of some job.
   * @return a boolean to signify if the jobName is or is not blacklisted or whitelisted.
   */
  public static boolean isJobTracked(final String jobName) {
    if ( DatadogUtilities.isJobBlacklisted(jobName) ) {
      return false;
    }
    return DatadogUtilities.isJobWhitelisted(jobName);
  }

  /**
   * Checks if a jobName is blacklisted.
   *
   * @param jobName - A String containing the name of some job.
   * @return a boolean to signify if the jobName is or is not blacklisted.
   */
  public static boolean isJobBlacklisted(final String jobName) {
    final String[] blacklist = DatadogUtilities.joblistStringtoArray( DatadogUtilities.getBlacklist() );
    final String jobNameLowerCase = jobName.toLowerCase();

    if (blacklist != null) {
      return Arrays.asList(blacklist).contains(jobNameLowerCase);
    }

    return false;
  }

  /**
   * Checks if a jobName is whitelisted.
   *
   * @param jobName - A String containing the name of some job.
   * @return a boolean to signify if the jobName is or is not whitelisted.
   */
  public static boolean isJobWhitelisted(final String jobName) {
    final String[] whitelist = DatadogUtilities.joblistStringtoArray( DatadogUtilities.getWhitelist() );
    final String jobNameLowerCase = jobName.toLowerCase();

    if ( whitelist != null ) {
      if ( whitelist.length == 0 ) {
        return true;
      }
      return Arrays.asList(whitelist).contains(jobNameLowerCase);
    }
    return true;
  }

  /**
   * Converts a blacklist/whitelist string into a String array.
   *
   * @param joblist - A String containing a set of job names.
   * @return a String array representing the job names to be blacklisted. Returns
   *         empty string if blacklist is null.
   */
  private static String[] joblistStringtoArray(final String joblist) {
    if ( joblist != null ) {
      String[] jobArr = joblist.split(",");
      if ( jobArr[0] != "" ) {
        return joblist.split(",");
      }
    }
    return ( new String[0] );
  }
```

src/main/resources/org/datadog/jenkins/plugins/datadog/DatadogBuildListener/global.jelly
```diff
@@ -23,9 +23,13 @@
     <f:validateButton title="${%Test Hostname}" progress="${%Testing...}"
                       method="testHostname" with="hostname" />
     <f:entry title="Blacklisted Jobs"
-             description="A comma-separated list of job names that should not monitored. (eg: susans-job,johns-job,prod-release)." >
+             description="A comma-separated list of job names that should not monitored. (e.g.: susans-job,johns-job,prod-release)." >
```
JohnLZeller: Let's add to this description something like `This takes precedent over whitelist.`

JohnLZeller:
Hey @bhavanki, just checking in with you again. Were you at all able to look at the feedback I gave?

sjenriquez:
@bhavanki closing this PR in favor of #78, thanks for your work on this! I'm preparing a release that should be out next week.

bhavanki:
@sjenriquez @JohnLZeller Thanks!

PR #73, opened 2016-12-13
Stop statsd client after sending data

suxor42:
The current implementation creates a new DescriptorImpl instance for every call
to onCompleted. For every new instance a new statsd client is created.
The statsd client is never stopped. This causes a resource(thread) leak.
The resource leak will cause a `java.lang.OutOfMemoryError: Unable to create new native thread`
How fast this happens depends on how often jobs are executed and the OS
https://plumbr.eu/outofmemoryerror/unable-to-create-new-native-thread#example

This PR stops the statsd client after it sent the data. A better fix would be to implement
DescriptorImpl as a real singleton


alexouzounis:
:+1 being hit by this, makes the plugin unusable

jshirley:
Same :( --  @JohnLZeller do you have any idea when this could be merged and/or if there is more we should do here?

jshirley-stripe:
Bump!

irabinovitch:
@jshirley Thank you for following up. We are moving closer to a release, as we have several PRs open.  Let me circle up with @JohnLZeller and see if we can offer a specific timeline.

PR #81, opened 2017-03-13
Add a gauge to measure build waiting time.

bbeck:
With a dynamically sized pool of agents it's important to know if you're scaling
appropriately and where you need to add more capacity.  This PR adds a metric for
each build to let you know how long it had to wait in the queue before executing.


src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -377,10 +377,11 @@ public static String nullSafeGetString(final JSONObject data, final String key)
    * @return a JSONArray containing a specific subset of tags retrieved from a builds metadata.
    */
   public static JSONArray assembleTags(final JSONObject builddata, final HashMap<String,String> extra) {
+    DatadogBuildListener.DescriptorImpl descriptor = DatadogUtilities.getDatadogDescriptor();
     JSONArray tags = new JSONArray();
 
     tags.add("job:" + builddata.get("job"));
-    if ( (builddata.get("node") != null) && DatadogUtilities.getDatadogDescriptor().getTagNode() ) {
+    if ( (builddata.get("node") != null) && descriptor.getTagNode() != null && descriptor.getTagNode() ) {
```
bbeck: This was throwing a NPE upon completion of every build if you were running Jenkins with the Datadog plugin installed but not configured.

The problem was the `tagNode` variable (in DatadogBuildListener) is null, but the `getTagNode` method tries to coerce it to a Boolean and fails.  Another possible way to fix this could be tomake `getTagNode` detect this case and just return `false` instead.  I'm happy to make this change if you feel that it's better.

sjenriquez: Thank for pointing this out! I made PR #84 to set `tagNode` to `false` by default, this will get merged soon. Could you keep this method as is and rebase once #84 is merged?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -97,16 +99,27 @@ public final void onStarted(final Run run, final TaskListener listener) {
...
 
+      HashMap<String,String> extraTags = new HashMap<String, String>();
+      try {
+        extraTags = DatadogUtilities.parseTagList(run, listener);
+      } catch (IOException ex) {
+        logger.severe(ex.getMessage());
+      } catch (InterruptedException ex) {
+        logger.severe(ex.getMessage());
+      }
+
```
sjenriquez: I think we should abstract this to a new method in `DataUtilities.java` so we can use it for future metrics as well. I was playing around with this and added this method:

```
  /**
   * Builds extraTags if any are configured in the Job
   *
   * @param run - Current build
   * @param listener - Current listener
   * @return A {@link HashMap} containing the key,value pairs of tags if any.
   */
  public static HashMap<String,String> buildExtraTags(Run run, TaskListener listener) {
    HashMap<String,String> extraTags = new HashMap<String, String>();
    try {
      extraTags = DatadogUtilities.parseTagList(run, listener);
    } catch (IOException ex) {
      logger.severe(ex.getMessage());
    } catch (InterruptedException ex) {
      logger.severe(ex.getMessage());
    }
    return extraTags;
  }
```

Then we can just add these lines here and on [line 160](https://github.com/DataDog/jenkins-datadog-plugin/pull/81/files/7b70f7995fd732377ae4f7bd6eb4997cfd1f7b4d?diff=split#diff-532720f1b459de301183a6478a8e6fa5R160) as well:
```
      HashMap<String,String> extraTags = new HashMap<String, String>();
      extraTags = DatadogUtilities.buildExtraTags(run, listener);
```

bbeck: Sure.  Just pushed an update that introduces this helper like you suggested.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
bbeck: Sure, no problem.  It's removed in the most recent update.

sjenriquez:
Thanks @bbeck, this looks good to go. Can you just rebase on master to resolve the merge conflict?

bbeck:
@sjenriquez Resolved the merge conflict and squashed commits.  It does look however like some tests are failing possibly related to this PR -- it seems that `Jenkins.getInstance()` is returning `null` when the test is initialized?  Have you seen anything like this before?  These tests were passing when I first submitted the PR, so maybe something has changed since?

sjenriquez:
@bbeck Sorry for the troubles here, indeed a PR was merged that broke the tests. This has been resolved and we've set up Travis-CI to avoid these issues in the future. Could you please resolve the  conflicts and push one more time. Thanks!

bbeck:
@sjenriquez Sorry for the delay.  I resolved conflicts, but the tests are still failing for what seems to be the same reason.  `Jenkins.getInstance()` is returning `null` when called from `Queue.getInstance()`.  I'm unsure of how to fix this -- this test used to run successfully.  Any ideas?

```
Caused by: java.lang.NullPointerException
	at hudson.model.Queue.getInstance(Queue.java:2856)
	at org.datadog.jenkins.plugins.datadog.DatadogBuildListener.<clinit>(DatadogBuildListener.java:81)
	... 38 more
```

sjenriquez:
Hey @bbeck, looks like we need some additional mocking. More info [available here](https://wiki.jenkins.io/display/JENKINS/Mocking+in+Unit+Tests). Check out the diff, tests are ✅  for me with these changes.

```
 $ git diff
diff --git a/src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java b/src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
index 7b10e5b..bff992f 100644
--- a/src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
+++ b/src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
@@ -8,9 +8,11 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
 import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
+import jenkins.model.Jenkins;

 import java.io.IOException;
 import java.util.HashMap;
@@ -22,13 +24,19 @@ import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.*;

 @RunWith(PowerMockRunner.class)
-@PrepareForTest({DatadogHttpRequests.class, DatadogUtilities.class})
+@PrepareForTest({DatadogHttpRequests.class, DatadogUtilities.class, Jenkins.class})
 public class DatadogBuildListenerTest {

+    @Mock
+    private Jenkins jenkins;
+
     private DatadogBuildListener datadogBuildListener;

     @Before
     public void setUp() throws Exception {
+        PowerMockito.mockStatic(Jenkins.class);
+        PowerMockito.when(Jenkins.getInstance()).thenReturn(jenkins);
+
         datadogBuildListener = spy(new DatadogBuildListener());

         PowerMockito.mockStatic(DatadogUtilities.class);
```

sjenriquez:
Thanks @bbeck !

PR #97, opened 2017-07-07
Use fullName instead of fullDisplayName for compatibility with multibranch workflow

### LINKED ISSUE #85: When used with GitHub org folder plugin; all jobs labeled master
When used in combination with GitHub Organization Folder Plugin, a lot of jobs are tagged as `#job:master` in datadog. This is due to the folder structure set up by this plugin (GitHubOrganizationName>ProjectName>BranchName).

This makes it hard to distinguish and collect metrics on specific projects in a github organization.

I'm thinking that a feasible approach would be adding another tag to jobs that specify parent folder or something similar. So that it won't interfere with regular nested folders.

lattwood:
Closes #85


joneshf:
This would fix a problem we're currently having. Any update on the status of this PR?

joneshf:
Woo! That's great. Thanks!

PR #117, opened 2018-06-18
Add support for global tags (including regexes)

nmuesch:
Adds a new textarea in the plugin that allows for the addition for the placement of tags globally for jobs. This supports regex and using capture groups for your tags. 

Ex: https://cl.ly/3f3r2h0R1R3q

Additional Notes:
I cleaned up the code a little bit and made the type of some objects the interface instead of the subclass type. 

Additionally right now, we iterate and compile the list of jobs on each run. I think ultimately it would be best to create some singleton class that holds the configuration objects and to only compile the list of regexes when the form is re saved/re configured. That seems outside the scope of this PR and there doesn't appear to be a noticeable performance hit.


src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -94,18 +96,22 @@ public DatadogBuildListener() { }
...
     String jobName = run.getParent().getFullName();
-    logger.fine(String.format("onStarted() called with jobName: %s", jobName));
-    HashMap<String,String> tags = new HashMap<String,String>();
+    Map<String,String> tags = new HashMap<String,String>();
 
     // Process only if job is NOT in blacklist and is in whitelist
     if ( DatadogUtilities.isJobTracked(jobName) ) {
       logger.fine("Started build!");
+
+      Boolean useJobRegex = DatadogUtilities.getDatadogDescriptor().getJobRegex();
```
gzussa: Change Boolean to its primitive. This will avoid creating a new object. Also, do we have to check for null here?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -151,8 +155,13 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
       logger.fine("Completed build!");
 
       // Collect Data
+      DatadogBuildListener.DescriptorImpl descriptor =  DatadogUtilities.getDatadogDescriptor();
+      Boolean useJobRegex = descriptor.getJobRegex();
```
gzussa: ditto

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -583,6 +594,13 @@ public boolean configure(final StaplerRequest req, final JSONObject formData)
         this.setTagNode(false);
       }
 
+      // Grab jobRegex and coerse to a boolean
+      if ( formData.getString("jobRegex").equals("true") ) {
```
gzussa: in case there is a null, you should do `"true".equals(formData.getString("jobRegex"))`

gzussa: what about `this.setJobRegex(formData.getString("jobRegex").equals("true"))`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -684,6 +702,26 @@ public void setWhitelist(final String jobs) {
       this.whitelist = jobs;
     }
 
+    /**
+     * Setter function for the {@link jobRegex} global configuration,
+     * accepting a boolean value (checkbox)
+     *
+     * @param jobRegex - a boolean flag on whether to use regexes instead of literals for whitelist/blacklisting jobs.
+     */
+    public void setJobRegex(final Boolean jobRegex) {
```
gzussa: why do we need `final`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogSCMListener.java
```diff
@@ -49,6 +49,11 @@ public void onCheckout(Run<?, ?> build, SCM scm, FilePath workspace, TaskListene
             && prop != null && prop.isEmitOnCheckout() ) {
       logger.fine("Checkout! in onCheckout()");
 
+      Boolean useJobRegex = DatadogUtilities.getDatadogDescriptor().getJobRegex();
```
gzussa: ditto

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -108,16 +107,64 @@ public static boolean isJobTracked(final String jobName) {
     return !DatadogUtilities.isJobBlacklisted(jobName) && DatadogUtilities.isJobWhitelisted(jobName);
   }
 
+  /**
+   * Retrieve the list of tags from the Config file if regex Jobs was checked
+   *  @param jobName - A string containing the name of some job
+   *  @return - A Map of values containing the key and value of each Datadog tag to apply to the metric/event
+   */
+  public static Map<String,String> getRegexJobTags(final String jobName) {
```
gzussa: I don't think `final` is needed here either. `final` would be needed if there would be an interface defining the method signature as such. `final` only prevent for params to be reassigned within the method itself and only for its scope. Thus, it's only really useful when defined in interfaces.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -108,16 +107,64 @@ public static boolean isJobTracked(final String jobName) {
...
+      Pattern p = Pattern.compile(jobInfo.get(0));
+      Matcher m = p.matcher(jobName);
+      if(m.matches()) {
+        for(int i = 1; i < jobInfo.size(); i++) {
+          String[] tagItem = jobInfo.get(i).split(":");
+          if(Character.toString(tagItem[1].charAt(0)).equals("$")) {
+            try {
+              tags.put(tagItem[0], m.group(Character.getNumericValue(tagItem[1].charAt(1))));
+            } catch(IndexOutOfBoundsException e) {
+              logger.fine(String.format("Specified a capture group that doesn't exist, not applying tag: " + Arrays.toString(tagItem) + " " + e));
```
gzussa: Could use `StringBuffer` here

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -108,16 +107,64 @@ public static boolean isJobTracked(final String jobName) {
...
    * Checks if a jobName is blacklisted.
    *
    * @param jobName - A String containing the name of some job.
    * @return a boolean to signify if the jobName is or is not blacklisted.
    */
-  public static boolean isJobBlacklisted(final String jobName) {
-    final List<String> blacklist = DatadogUtilities.joblistStringtoList( DatadogUtilities.getBlacklist() );
-
-    return blacklist.contains(jobName.toLowerCase());
+  public static boolean  isJobBlacklisted(final String jobName) {
```
gzussa: ditto and everywhere else :)

src/main/java/org/datadog/jenkins/plugins/datadog/BuildStartedEventImpl.java
```diff
@@ -11,9 +13,9 @@
 public class BuildStartedEventImpl implements DatadogEvent  {
 
   private JSONObject builddata;
-  private HashMap<String,String> tags;
+  private Map<String,String> tags;
```
gzussa: 👍

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
nmuesch: After talking with @sjenriquez about the scope/UI complexity here we decided to move to a new text area for the tags option. In doing this I removed the need for this variable entirely. Good to know though!

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
nmuesch: There should be a non trivial performance bump by changing all of the `String.format` to something like StringBuilder or StringBuffer. We should change all occurrences in a separate PR though as it seems a little out of scope here.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -703,23 +696,23 @@ public void setWhitelist(final String jobs) {
     }
 
     /**
-     * Setter function for the {@link jobRegex} global configuration,
-     * accepting a boolean value (checkbox)
+     * Getter function for the {@link globalJobTags} global configuration, containing
```
gzussa: 👍 nice! didn't even saw that. Good catch 😸

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -576,6 +584,9 @@ public boolean configure(final StaplerRequest req, final JSONObject formData)
       // Grab whitelist
       this.setWhitelist(formData.getString("whitelist"));
 
+      // Grab the Global Job Tags
+      this.setGlobalJobTags(formData.getString("globalJobTags"));
+
       // Grab tagNode and coerse to a boolean
       if ( formData.getString("tagNode").equals("true") ) {
```
gzussa: in case there is a null, you should do "true".equals(formData.getString("jobRegex"))

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
gzussa: ok but then either you use `format` or `+` sign concatenation but both doesn't make sense to me.

nmuesch:
Thanks for all the feedback. I agree that there should be a followup cleaning PR for a couple of the brought up points 🙇

PR #118, opened 2018-08-03
Add node tag to the jenkins.job.waiting

keirbadger:
Added so the metric can be split in Datadog to show the wait time
per node


sjenriquez:
@keirbadger Thank you for the PR! I'm going to include this change in our upcoming release. In order to get this out asap I opened a new PR, #119 that doesn't touch the `pom.xml` file. Thanks again!

PR #121, opened 2018-09-05
implemented new TestCompletedEvent

jaimecasero:


jaimecasero:
@sjenriquez ups sorry, I didnt mean to submit PR on original fork yet. 

I was just elaborating an experiment to report datadog events for JenkinsJobs failed tests. 

These tests in our solution are not regular unit tests, they are more E2E tests run against production environment, so we want these test results to be triggered on DataDog in case they fail.

It would convenient to alert based on the testcases, rather than create a Job per TestCase.

So this PR is about triggering DataDog events from test failures, instead of plain Job status

PR #132, opened 2018-11-19
Send delivery KPIs (fixes #130)

### LINKED ISSUE #130: Have pipeline statistics
I'd like to see pipeline statistics as mentioned in [this blog post](https://stelligent.com/2017/11/16/codepipeline-dashboard/)

Each `Job` should provide
- Lead Time
- Cycle Time
- Feedback Time
- MTTR
- MTBF

Fail 1

From [https://github.com/stelligent/pipeline-dashboard/blob/master/README.md#metric-details](https://github.com/stelligent/pipeline-dashboard/blob/master/README.md#metric-details)

What do you think, is it worth to start a PR?

pgarbe:
This PRs sends additional metrics to Datadog to measure the performance of teams. It's based on the  [blog post](https://stelligent.com/2017/11/16/codepipeline-dashboard/) from Stelligent.

Each `Job`  provides
- Lead Time
- Cycle Time
- Feedback Time
- MTTR
- MTBF

Fail 1

From [https://github.com/stelligent/pipeline-dashboard/blob/master/README.md#metric-details](https://github.com/stelligent/pipeline-dashboard/blob/master/README.md#metric-details)

### DIFF
--- a/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
+++ b/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
@@ -11,6 +11,7 @@ import hudson.model.AbstractProject;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
 import hudson.model.Queue;
+import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.model.listeners.RunListener;
@@ -181,9 +182,9 @@ public class DatadogBuildListener extends RunListener<Run>
       }
 
       // Setup tags for StatsDClient reporting
-      String[] tagsToCounter = new String[tagArr.size()];
+      String[] statsdTags = new String[tagArr.size()];
       for(int i = 0; i < tagArr.size(); i++) {
-        tagsToCounter[i] = tagArr.getString(i);
+        statsdTags[i] = tagArr.getString(i);
       }
 
       // Report to StatsDClient
@@ -194,8 +195,35 @@ public class DatadogBuildListener extends RunListener<Run>
           // The client is a threadpool so instead of creating a new instance of the pool
           // we lease the exiting one registerd with Jenkins.
           statsd = getDescriptor().leaseClient();
-          statsd.incrementCounter("completed", tagsToCounter);
-          logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(tagsToCounter)));
+          statsd.incrementCounter("completed", statsdTags);
+          logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(statsdTags)));
+
+          // Send KPIs 
+          if (!isMasterBranch(run)) {
+            if (run.getResult() == Result.SUCCESS) {
+              long mttr = getMeanTimeToRecovery(run);
+              long cycleTime = getCycleTime(run);;
+              long leadTime = run.getDuration() + mttr;
+
+              statsd.gauge("leadtime", leadTime / 1000, statsdTags);
+              if (cycleTime > 0) {
+                statsd.gauge("cycletime", cycleTime / 1000, statsdTags);
+              }
+              if (mttr > 0) {
+                statsd.gauge("mttr", mttr / 1000, statsdTags);
+              }
+            } else {
+              long feedbackTime = run.getDuration();
+              long mtbf = getMeanTimeBetweenFailure(run);
+
+              statsd.gauge("feedbacktime", feedbackTime / 1000, statsdTags);
+
+              if (mtbf > 0) {
+                statsd.gauge("mtbf", mtbf / 1000, statsdTags);
+              }
+            }
+          }
+
         } catch (StatsDClientException e) {
           logger.severe(String.format("Runtime exception thrown using the StatsDClient. Exception: %s", e.getMessage()));
         } finally {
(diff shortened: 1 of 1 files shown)


miketheman:
Looks awesome!! @sjenriquez any chance for a review soon?

sjenriquez:
@pgarbe Thanks for the PR, will take a look shortly! cc: @miketheman

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -196,6 +197,26 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
           statsd = getDescriptor().leaseClient();
           statsd.incrementCounter("completed", tagsToCounter);
           logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(tagsToCounter)));
+
+          // Send KPIs 
+          if (!isMasterBranch(run)) {
```
sjenriquez: Why do we only want to collect metrics for `!isMasterBranch(run)`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -196,6 +197,26 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
+
+              statsd.gauge("cycletime", cycleTime / 1000, tagsToCounter);
+              statsd.gauge("leadtime", leadTime / 1000, tagsToCounter);
+              statsd.gauge("mttr", mttr / 1000, tagsToCounter);
+            } else {
+              long feedbackTime = run.getDuration();
+              long mtbf = getMeanTimeBetweenFailure(run);
+
+              statsd.gauge("feedbacktime", feedbackTime / 1000, tagsToCounter);
+              statsd.gauge("mtbf", mtbf / 1000, tagsToCounter);
```
sjenriquez: Submitting `0` values could lead to incorrect aggregation over various timeframes. In these cases, we should only submit the metric if the value `!= 0`. Looks like this could be the case for `mttr`, `cycleTime`, and `mtbf`.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -196,6 +197,26 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
           logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(tagsToCounter)));
+
+          // Send KPIs 
+          if (!isMasterBranch(run)) {
+            if (run.getResult() == Result.SUCCESS) {
+              long mttr = getMeanTimeToRecovery(run);
+              long cycleTime = getCycleTime(run);;
+              long leadTime = run.getDuration() + mttr;
+
+              statsd.gauge("cycletime", cycleTime / 1000, tagsToCounter);
```
sjenriquez: Small thing: can you rename `tagsToCounter` to something like `statsdTags`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
pgarbe: These metrics make only sense when it's about a release. And my assumption is that releases happens only on the master branch. That's why it doesn't make sense to consider PR or branch builds.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
pgarbe: Yes, that makes sense. Will change it.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
pgarbe: No problem

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
sjenriquez: Yea, I see how this is tied to code releases. We could hit issues trying to match a hardcoded `master` branch that could really limits these metrics. It's possible releases happen on branches not named `master` or some other use case with a name different from `master`. I think we'll need some options in the Job configuration page to make this flexible - a checkbox to enable/disable collection of these metrics as well as an option to specify a branch name. What are your thoughts?

miketheman: How about instead of making the branch a configuration value, tracking _all_ branches, and adding that as a tag, so that the user can then decide which ones they want to use for overlays, metrics, etc?

sjenriquez: 💯 @miketheman

pgarbe: If the branch should be added as tag, it would affect also all other metrics. Do you think that make sense?

miketheman: Maybe the branch tag should be associated with every metric, so as to allow for better analysis

sjenriquez: Apologies for not catching this earlier, but we already collect the branch tag if it exists: https://github.com/DataDog/jenkins-datadog-plugin/blob/27fc3a58014aa4a84a9c35173b7f8d254355c5e2/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java#L314-L315

pgarbe: Thanks for the hint. Actually, I've never seen this tag in one of our metrics. Have to check why this is missing.

srizzling:
How close is this to hitting master?

irabinovitch:
@sjenriquez @pgarbe Thoughts on next steps?

sjenriquez:
@irabinovitch This PR is mostly there, we can finish it up internally and get it merged.

nmuesch:
Thanks again for this @pgarbe! I've done the requested cleanups and merged this in #132 🙇

PR #138, opened 2019-01-10
Search path for hostname executable

shonfeder-da:
The first tiny commit was added on the way to second tiny one. 

For context, without the change to the docs, when I ran `mvn package` I would get

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.10.1:javadoc (default) on project datadog: An error has occurred in JavaDocs report generation:
[ERROR] Exit code: 1 - /Users/shonfeder/oss/jenkins-plugins/datadog-plugin-jenkinsci/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java:386: error: tag not supported in the generated HTML version: tt
[ERROR]    * <p>See <tt>DatadogBuildListener/*.jelly</tt> for the actual HTML fragment
```

We've tested to confirm that calling hostname via `/usr/bin/env` resolves the issue we were facing.

### DIFF
--- a/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
+++ b/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
@@ -383,7 +383,7 @@ public class DatadogBuildListener extends RunListener<Run>
    * Descriptor for {@link DatadogBuildListener}. Used as a singleton.
    * The class is marked as public so that it can be accessed from views.
    *
-   * <p>See <tt>DatadogBuildListener/*.jelly</tt> for the actual HTML fragment
+   * <p>See <code>DatadogBuildListener/*.jelly</code> for the actual HTML fragment
    * for the configuration screen.
    */
   @Extension // Indicates to Jenkins that this is an extension point implementation.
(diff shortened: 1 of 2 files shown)


shonfeder-da:
@sjenriquez Any interest in merging this in? If not, I'll go ahead and close it. We no longer have a need for this fix (and we are using a patched version any how).

sjenriquez:
Hi @shonfeder-da, sorry for the delay in my response. I'm hesitant to merge to prevent causing any new issues for existing installs. It sounds like you've resolved the issue, did you try reinstalling the package? I'm going to close the PR but please open an issue if you have any continued problems, thanks.

shonfeder-da:
No worries!

What kind of issues do you think it would cause for existing installs?

To be clear: we resolved the issue by creating an internal fork of the plugin that patched this hard-coded path to `hostname` with the proper delegation to the environment. From our perspective, the plugin is broken on our distro of choice, since it makes unnecessary assumptions about the file structure.

flokli:
I'm not sure how this should break existing installs.

`/usr/bin/env hostname` will take the first `hostname` binary from `$PATH` (while we can assume `/bin` to be in `$PATH`, and whatever `hostname` is first in `$PATH` to be not terribly broken, otherwise you'll have more serious problems)

The current `/bin/hostname` will only work if exactly `/bin/hostname` is present - something that is especially not true on systems like NixOS and Guix - what this PR would fix.

See https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

PR #150, opened 2019-04-09
Add a timeout to HTTP request connection and reads

Mischa-Alff:
This prevents job hangs when the connection to datadog is flaky. Currently set to 3 minutes

### DIFF
--- a/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogHttpRequests.java
+++ b/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogHttpRequests.java
@@ -55,6 +55,14 @@ public class DatadogHttpRequests {
       logger.fine("Using HttpURLConnection, without proxy");
     }
 
+    /* Timeout of 3 minutes for connecting and reading.
+     * this prevents this plugin from causing jobs to hang in case of
+     * flaky network or Datadog being down. Left intentionally long.
+     */
+    int timeoutMS = 3 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


azzeddinefaik:
nice work here! I really need this timeout functionality! Thanks @Mischa-Alff

nmuesch:
Hey @Mischa-Alff Thanks for the contribution 🙇 Do you think this would be helpful to be made configurable?

Mischa-Alff:
Hey! What would the use case for being able to configure this be?

nmuesch:
Three minutes, while reasonable, seems a little arbitrary is all 😄 Just was wondering if you had thought it was worth it being user defined with a three minute default. I don't have a strong opinion on either option.

Mischa-Alff:
I agree that it's arbitrary, I'm not sure the user cares about such an implementation detail though.

dabcoder:
@Mischa-Alff sorry for the delay on this one. Could you merge or rebase your branch against the master one? There has been some changes and the `DatadogHttpRequests.java` no longer exists, so the addition you want to make should be in [DatadogHttpClient.java](https://github.com/DataDog/jenkins-datadog-plugin/blob/master/src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java) I believe.

gzussa:
Closing this pull request since I am porting this change in https://github.com/DataDog/jenkins-datadog-plugin/pull/174. Thanks @Mischa-Alff

PR #156, opened 2019-06-01
Send Delivery KPIs

nmuesch:
Takes the work of #132 and adds on top of it. 

Since 132 was a fork I couldn't add commits directly to it, I've made this new PR but was able to keep original commit authorship here.

### DIFF
--- a/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
+++ b/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
@@ -108,7 +108,7 @@ public class DatadogBuildListener extends RunListener<Run>
 
       // Get the list of global tags to apply
       tags.putAll(DatadogUtilities.getRegexJobTags(jobName));
-  
+
       Queue.Item item = queue.getItem(run.getQueueId());
 
       // Gather pre-build metadata
@@ -204,29 +204,28 @@ public class DatadogBuildListener extends RunListener<Run>
           statsd.incrementCounter("completed", statsdTags);
           logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(statsdTags)));
 
-          // Send KPIs 
-          if (!isMasterBranch(run)) {
-            if (run.getResult() == Result.SUCCESS) {
-              long mttr = getMeanTimeToRecovery(run);
-              long cycleTime = getCycleTime(run);;
-              long leadTime = run.getDuration() + mttr;
+          logger.fine("Computing KPI metrics");
+          // Send KPIs
+          if (run.getResult() == Result.SUCCESS) {
+            long mttr = getMeanTimeToRecovery(run);
+            long cycleTime = getCycleTime(run);;
+            long leadTime = run.getDuration() + mttr;
 
-              statsd.gauge("leadtime", leadTime / 1000, statsdTags);
-              if (cycleTime > 0) {
-                statsd.gauge("cycletime", cycleTime / 1000, statsdTags);
-              }
-              if (mttr > 0) {
-                statsd.gauge("mttr", mttr / 1000, statsdTags);
-              }
-            } else {
-              long feedbackTime = run.getDuration();
-              long mtbf = getMeanTimeBetweenFailure(run);
+            statsd.gauge("leadtime", leadTime / 1000, statsdTags);
+            if (cycleTime > 0) {
+              statsd.gauge("cycletime", cycleTime / 1000, statsdTags);
+            }
+            if (mttr > 0) {
+              statsd.gauge("mttr", mttr / 1000, statsdTags);
+            }
+          } else {
+            long feedbackTime = run.getDuration();
+            long mtbf = getMeanTimeBetweenFailure(run);
 
-              statsd.gauge("feedbacktime", feedbackTime / 1000, statsdTags);
+            statsd.gauge("feedbacktime", feedbackTime / 1000, statsdTags);
 
-              if (mtbf > 0) {
-                statsd.gauge("mtbf", mtbf / 1000, statsdTags);
-              }
+            if (mtbf > 0) {
+              statsd.gauge("mtbf", mtbf / 1000, statsdTags);
             }
           }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -200,8 +201,34 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
-          statsd.incrementCounter("completed", tagsToCounter);
-          logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(tagsToCounter)));
+          statsd.incrementCounter("completed", statsdTags);
+          logger.fine(String.format("Attempted to send 'completed' counter with tags: %s", Arrays.toString(statsdTags)));
+
+          logger.fine("Computing KPI metrics");
+          // Send KPIs
+          if (run.getResult() == Result.SUCCESS) {
+            long mttr = getMeanTimeToRecovery(run);
+            long cycleTime = getCycleTime(run);;
```
gzussa: extra `;`

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -200,8 +201,34 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
+            long leadTime = run.getDuration() + mttr;
+
+            statsd.gauge("leadtime", leadTime / 1000, statsdTags);
```
gzussa: Let's use a constant for `1000`

gzussa: Shouldn't the name be prefixed with `jenkins.job`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -200,8 +201,34 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
+            if (cycleTime > 0) {
```
gzussa: Why don't we report if `0`?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -200,8 +201,34 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
+              statsd.gauge("cycletime", cycleTime / 1000, statsdTags);
+            }
+            if (mttr > 0) {
```
gzussa: Same as above and below

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -221,7 +248,40 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
       logger.fine("Finished onCompleted()");
     }
   }
-  
+
+  private long getMeanTimeBetweenFailure(final Run run) {
```
gzussa: `final` doesn't seem to be needed here (same comment below)

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -221,7 +248,40 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
+              (previousSuccessfulBuild.getStartTimeInMillis() + previousSuccessfulBuild.getDuration());
+    }
+    return 0;
+  }
+
+  private long getMeanTimeToRecovery(final Run run) {
+    if (buildFailed(run.getPreviousBuiltBuild())) {
+      Run<?, ?> firstFailedRun = run.getPreviousBuiltBuild();
+
+      while (buildFailed(firstFailedRun.getPreviousBuiltBuild())) {
```
gzussa: Could this lead to an infinite loop? Should we have a max just in case?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
nmuesch: Good 🦅 👁 catch!

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
nmuesch: When we lease the statsd client, we setup the prefix then - https://github.com/DataDog/jenkins-datadog-plugin/blob/46b025d16bb1b8725664aa0f822a77ded5b2d592/src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java#L404

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
nmuesch: Its because the functions that calculate these values use 0 as a "nothing to report" kind of special value. Like if there is no previous run that was successful, this metric doesn't make sense, and the function that calculates this value would return a value that doesn't really mean anything. Seems like the same for below.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
nmuesch: See my note above.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
nmuesch: I took a look at the Java docs (as I had a similar thought) and my read was that we'd stop at the root if there were no failed runs (which seems a bit unlikely to me though)

gzussa: What about performance, could it be that it loops for so long that it makes the all thing slow. Maybe too far fetched. You tell me.

PR #165, opened 2019-09-30
replace travis with azure pipelines

gzussa:


.azure-pipelines/azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
...
+- task: Maven@3
+  inputs:
+    mavenPomFile: 'pom.xml'
+    mavenOptions: '-Xmx3072m'
+    javaHomeOption: 'JDKVersion'
+    jdkVersionOption: '1.8'
+    jdkArchitectureOption: 'x64'
+    publishJUnitResults: true
+    testResultsFiles: '**/surefire-reports/TEST-*.xml'
+    goals: 'package'
```
nmuesch: should this be `test`?

.azure-pipelines/azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
...
+    #mavenDirectory: # Required when mavenVersionOption == Path
+    #mavenSetM2Home: false # Required when mavenVersionOption == Path
+    #mavenOptions: '-Xmx1024m' # Optional
+    #mavenAuthenticateFeed: false
+    #effectivePomSkip: false
+    #sonarQubeRunAnalysis: false
+    #sqMavenPluginVersionChoice: 'latest' # Required when sonarQubeRunAnalysis == True# Options: latest, pom
+    #checkStyleRunAnalysis: false # Optional
+    #pmdRunAnalysis: false # Optional
+    #findBugsRunAnalysis: false # Optional
```
nmuesch: Can we remove the commented out options and add them later if needed?

.azure-pipelines/azure-pipelines.yml
```diff
@@ -0,0 +1,32 @@
...
+
+steps:
+- task: Maven@3
+  inputs:
+    mavenPomFile: 'pom.xml'
+    mavenOptions: '-Xmx3072m'
+    javaHomeOption: 'JDKVersion'
+    jdkVersionOption: '1.8'
+    jdkArchitectureOption: 'x64'
+    publishJUnitResults: true
```
nmuesch: whats this 😄

gzussa: see the `Publish to Azure Pipelines` item in https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/maven?view=azure-devops

.azure-pipelines/azure-pipelines.yml
gzussa: I don't think it does any harm to test the entire packaging.

nmuesch: Will this run the tests though? From the maven docs it looks like `package` just takes the compiled code and tries to bundle it into (in our case) a jar - http://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html

`mvn test` should handle both the compilation + testing step for us.

nmuesch: Ahh I see that in the azure-pipelines run that tests did indeed get run, the output is a bit difficult to read, but in that case I don't have strong opinions about the `goal` definition

Issue #34, opened 2016-01-08
Provide option to post to local dogstatsd rather than going directly to the API


gregsymons:
On occasion I've had issues with posts to the API timing out and then "failing" what would have otherwise been a successful build. Since I'm generally running the agent somewhere nearby Jenkins anyway, it'd be nice to leverage that to have a little safety against hiccups in API availability.

JohnLZeller:
Hey @gregsymons! Thanks for the ticket!

So I can understand your specific situation a bit better, could you clarify what happened? Your build "failed" when the plugin failed to post? Or, the plugin failed when it failed to post?

As far as posting to a dogstatsd somewhere. When we created the jenkins plugin, we deprecated a jenkins agent check. Introducing this option to send to a local dogstatsd could provide some safety, but I'm curious if it's the best option. I'm going to take a further look at this and get back to you here.

gregsymons:
Unfortunately, I don't have any build output to show exactly what was happening--I disabled the plugin to keep things moving, and all the failed builds have been rotated into dust by now. But, what I remember is that the build would stall for a long time after logging the following:

```
DatadogBuildListener.java: Completed build!
DatadogBuildListener.java: Using hostname set in 'Manage Plugins'. Hostname: ci.gwfathom.com
DatadogBuildListener.java: Sending event
```
 and then would eventually timeout and throw an exception, which would change the build result to `FAILURE` even though the build itself was otherwise successful. It _may_ have been an interaction with the Workflow plugin, since the builds in question are Workflow builds, but I don't have any direct evidence of that.

JohnLZeller:
@gregsymons I'm curious if you've run into more issues with this since your last comment. Could be a good use case to fire at DogStatsD if a none 200 response from the API occurs.

mccutchen:
+1 for being able to configure this plugin to publish only to the local `dogstatsd` instance.  I see a couple significant benefits to that approach:
 - Jenkins no longer needs to know any DataDog API keys (simpler plugin configuration & one fewer way to leak secrets)
 - Metrics will automatically pick up host-level global tags (simpler plugin configuration & automatic tagging consistency with all other metrics recorded elsewhere in the infrastructure)

JohnLZeller:
All valid points @mccutchen, I'll prioritize this. Thanks!

PR #169, opened 2019-11-14
Send all metrics w/o DogstatsD

dabcoder:
- Get rid of the DogStatsD dependency by creating a thread safe counter cache 
  - Use a ConcurrentMap to make sure that increments are thread safe - prevent concurrency issue - I may have been over protective with my implementation
  - Use a ReentrantLock to make sure we don't flush and increment the counter cache at the same time (across threads).
  - Use the singleton pattern to only have one instance of the cache in the JVM (also thread safe)
  - Counter are submitted as COUNT with an interval set to 10 s (recommended interval)
- It also adds `jenkins.job.started` and `jenkins.scm.checkout` metrics.
- It increases test coverage!
- Wrapping all extensions in global `try catch` statements so build don't fail because of the plugin - see issue discussed in https://github.com/DataDog/jenkins-datadog-plugin/issues/34 (however, in this PR we are removing DogStatsD dependency)
- Changed the DatadogHttpClient into another singleton - thread safe - Hence we don't have lease logic anymore from the descriptor.


pom.xml
```diff
@@ -99,6 +99,14 @@
         <artifactId>maven-site-plugin</artifactId>
         <version>3.5</version>
       </plugin>
+      <plugin>
```
gzussa: Do we need this?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogMetricCounter.java
```diff
@@ -0,0 +1,43 @@
...
+
+@Extension
+public class DatadogMetricCounter extends PeriodicWork {
+
+    private static final Logger logger = Logger.getLogger(DatadogMetricCounter.class.getName());
+
+    @Override
+    public long getRecurrencePeriod() {
+        // run frequency - 60 seconds
+        return PeriodicWork.MIN;
```
gzussa: Let's change this frequency to 10 or 15 sec

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogMetricCounter.java
```diff
@@ -0,0 +1,43 @@
...
+
+import hudson.Extension;
+import hudson.model.*;
+import org.datadog.jenkins.plugins.datadog.clients.DatadogHttpClient;
+import java.util.concurrent.ConcurrentMap;
+import java.util.logging.Logger;
+import org.datadog.jenkins.plugins.datadog.model.LocalCacheCounters;
+
+@Extension
+public class DatadogMetricCounter extends PeriodicWork {
```
gzussa: I am still not convinced about the name of this class. It submits counters. Maybe we should call this `PeriodicCounterForwarder` or something similar. Also, in order to be more generic we could change the logic a little bit so that we also specify the metric name for each count.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogMetricCounter.java
```diff
@@ -0,0 +1,43 @@
...
+        for (String tags: cache.keySet()) {
+            int counter = cache.get(tags);
+
+            ((DatadogHttpClient) client).gauge("jenkins.job.completed",
+                    counter,
+                    DatadogUtilities.getHostname(null),
+                    tags);
+        }
+
+        LocalCacheCounters.Cache.set(cache);
```
gzussa: The cache should be reinitialized here.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogMetricCounter.java
```diff
@@ -0,0 +1,43 @@
...
+    protected void doRun() throws Exception {
+        if (DatadogUtilities.isApiKeyNull()) {
+            return;
+        }
+
+        // Instantiate the Datadog Client
+        DatadogClient client = DatadogUtilities.getDatadogDescriptor().leaseDatadogClient();
+
+        ConcurrentMap<String, Integer> cache = LocalCacheCounters.Cache.get();
+
```
gzussa: Let's reset the cache here instead of at the end

src/main/java/org/datadog/jenkins/plugins/datadog/model/LocalCacheCounters.java
```diff
@@ -0,0 +1,14 @@
+package org.datadog.jenkins.plugins.datadog.model;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class LocalCacheCounters {
```
gzussa: Let's rename this into `ConcurrentMetricCounters`

src/main/java/org/datadog/jenkins/plugins/datadog/model/LocalCacheCounters.java
```diff
@@ -0,0 +1,14 @@
...
+    private static final ConcurrentMap<String, Integer> counters = new ConcurrentHashMap<String, Integer>();
+    public static final ThreadLocal<ConcurrentMap<String, Integer>> Cache =
```
gzussa: Let's rename this counters

src/main/java/org/datadog/jenkins/plugins/datadog/model/LocalCacheCounters.java
```diff
@@ -0,0 +1,14 @@
+package org.datadog.jenkins.plugins.datadog.model;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class LocalCacheCounters {
+    private static final ConcurrentMap<String, Integer> counters = new ConcurrentHashMap<String, Integer>();
```
gzussa: Let's rename this one metrics.
Also instead of a String as a key, we could also pass the metric name. So we should create a Class that would have a JSONArray for storing tags and a String for the metric name. Then we would have to implement the `equals` and `hashCode` methods so it works with HashMap. This way we could get rid of the `gauge` method we created.

pom.xml
dabcoder: This was added when moving from using the openjdk 9 to the oracle one (version 8). Not sure we need to keep this but without it the code didn't compile with `mvn`.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogPeriodicCounterSender.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Extension
+public class DatadogPeriodicCounterSender extends PeriodicWork {
+
+    private static final Logger logger = Logger.getLogger(DatadogPeriodicCounterSender.class.getName());
+
+    @Override
+    public long getRecurrencePeriod() {
+        // run frequency - 15 seconds
+        return PeriodicWork.MIN * 60 / 4;
```
dabcoder: This should be `return PeriodicWork.MIN / 4;` (every 15 seconds, not 15 minutes). Will update.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -690,23 +581,6 @@ public void setTargetMetricURL(String targetMetricURL) {
...
-                }
-            } catch (Exception e) {
-                logger.severe(String.format("Error while configuring StatsDClient. Exception: %s", e.toString()));
-            }
-            return statsDClient;
-        }
-
         /**
          * @return - A {@link DatadogClient} lease for this registered {@link RunListener}
          */
```
dabcoder: This also keeps logging `logger.warning("datadogClient is already set");` because we have `DatadogClient client = getDescriptor().leaseDatadogClient();` in 2 methods (`OnStarted` and `OnCompleted`).

src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
```diff
@@ -133,23 +165,24 @@ public void testOnCompletedWithDurationAsZero() throws Exception {
...
         expectedTags[2] = "result:SUCCESS";
         expectedTags[3] = "branch:test-branch";
         client.assertMetric("jenkins.job.duration", 0, "null", expectedTags);
+        client.assertMetric("jenkins.job.leadtime", 0, "null", expectedTags);
         client.assertServiceCheck("jenkins.job.status", 0, "null", expectedTags);
         client.assertedAllMetricsAndServiceChecks();
     }
 
     @Test
-    public void testOnCompletedSuccessWithStatsD() throws Exception {
```
gzussa: Indeed, this test is irrelevant now since we don't have StatsD

src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
```diff
@@ -133,23 +165,24 @@ public void testOnCompletedWithDurationAsZero() throws Exception {
...
+    public void testOnFailures() throws Exception {
```
gzussa: What value does this tests bring compare to the `testOnCompletedWithEverything` one?

src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
```diff
@@ -165,32 +198,56 @@ public void testOnCompletedSuccessWithStatsD() throws Exception {
...
+        // getPreviousBuiltBuild()
+        when(lastRun.getPreviousSuccessfulBuild()).thenReturn(previousRun);
+        Assert.assertNotNull(listener.getMeanTimeToRecovery(lastRun));
     }
 
-    private DatadogBuildListener.DescriptorImpl descriptor(DatadogClient client, StatsDClient statsd) {
+    private DatadogBuildListener.DescriptorImpl descriptor(DatadogClient client) {
         DatadogBuildListener.DescriptorImpl descriptor = mock(DatadogBuildListener.DescriptorImpl.class);
         when(descriptor.leaseDatadogClient()).thenReturn(client);
-        when(descriptor.leaseStatDClient()).thenReturn(statsd);
```
gzussa: 👍

src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
```diff
@@ -165,32 +198,56 @@ public void testOnCompletedSuccessWithStatsD() throws Exception {
...
+        client.assertMetric("jenkins.job.feedbacktime", 123, "null", expectedTags);
+        client.assertMetric("jenkins.job.mtbf", 123, "null", expectedTags);
+        client.assertServiceCheck("jenkins.job.status", 2, "null", expectedTags);
+    }
 
-        statsd.assertMetric("completed", 1, expectedTags);
-        statsd.assertMetric("leadtime", 1234, expectedTags);
-        statsd.assertedAllMetricsAndCounters();
+    @Test
+    public void testJenkinsRuns() throws Exception {
```
gzussa: I am not sure to understand what this test does.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogPeriodicCounterSender.java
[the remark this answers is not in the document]
dabcoder: Yes see: https://github.com/DataDog/jenkins-datadog-plugin/pull/169#discussion_r347378586

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -179,137 +185,82 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
...
+    public long getMeanTimeToRecovery(Run<?, ?> run) {
         if (buildFailed(run.getPreviousBuiltBuild())) {
             Run<?, ?> firstFailedRun = run.getPreviousBuiltBuild();
 
             while (firstFailedRun != null && buildFailed(firstFailedRun.getPreviousBuiltBuild())) {
                 firstFailedRun = firstFailedRun.getPreviousBuiltBuild();
             }
             if (firstFailedRun != null) {
-                return run.getStartTimeInMillis() - firstFailedRun.getStartTimeInMillis();
+                long mttf = run.getStartTimeInMillis() - firstFailedRun.getStartTimeInMillis();
```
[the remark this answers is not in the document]
dabcoder: I thought it would make the returned values more readable, but it looks like this is not the right way to do it. Reverting

src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
dabcoder: `testOnCompletedWithEverything` only tests a run that succeeds:  https://github.com/DataDog/jenkins-datadog-plugin/blob/fd3fe2f26a80f7b595f99c04e5e68bd8114496e7/src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java#L93. For `testOnFailures()`, we're testing 2 different metrics (`mtbf` and `feedbacktime`).

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListener.java
```diff
@@ -164,6 +160,16 @@ public final void onCompleted(final Run run, @Nonnull final TaskListener listene
                 buildData.getHostname("null"),
                 tags);
 
+        ConcurrentMap<CounterMetric, Integer> counters = ConcurrentMetricCounters.Counters.get();
+        System.out.println("Sending the jenkins.job.completed metric with " + counters.toString());
+        CounterMetric counterMetric = new CounterMetric(tags, "jenkins.job.completed");
+        if (counters.get(counterMetric) == null) {
```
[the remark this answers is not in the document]
dabcoder: Getting `Undefined reference: Object java.util.concurrent.ConcurrentMap.getOrDefault(Object, Object)` when compiling. Looks like this is due to a Java 1.7 -> 1.8 issue.

src/test/java/org/datadog/jenkins/plugins/datadog/DatadogBuildListenerTest.java
dabcoder: As per your comment above: 

> Also we should go deeper testing mttr, cycleTime, leadTime metrics. Each deserve their own tests.

This is to test that the values we are getting for the methods that generate these variables are not null.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogClient.java
```diff
@@ -21,6 +21,21 @@
...
+     * @param name - metric name
+     * @param hostname - metric hostname
+     * @param tags - metric tags
+     */
+    public void incrementCounter(String name, String hostname, JSONArray tags);
+
+    /**
+     * Submit all your counters as rate with 10 seconds intervals.
+     */
+    public void flushCounters();
```
zippolyte: Shouldn't this be a flushAll method, that would flush gauges and service checks and everything every 10s ?

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
```diff
@@ -55,8 +74,39 @@ public boolean sendEvent(JSONObject payload) {
...
+    @Override
+    public void flushCounters() {
+        ConcurrentMap<CounterMetric, Integer> counters = ConcurrentMetricCounters.Counters.get();
+        ConcurrentMetricCounters.Counters.set(new ConcurrentHashMap<CounterMetric, Integer>());
+
+        // Submit all metrics as gauge
+        for (CounterMetric counterMetric: counters.keySet()) {
+            int count = counters.get(counterMetric);
+
+            this.postMetric(counterMetric.getMetricName(), count, counterMetric.getHostname(), counterMetric.getTags(), "rate");
```
zippolyte: Why send as rate and not as count ?

zippolyte: I guess backward compatibility with what dogstatsd was doing before ?

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogClient.java
gzussa: Maybe we should do the same for metrics and services checks indeed. However I will not do it in this PR.

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
gzussa: I should have read this feedback sooner... Would have prevented some long debugging

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
```diff
@@ -2,13 +2,31 @@
...
+import java.util.concurrent.locks.ReentrantLock;
 
 public class ConcurrentMetricCounters {
-    private static final ConcurrentMap<CounterMetric, Integer> metrics = new ConcurrentHashMap<>();
-    public static final ThreadLocal<ConcurrentMap<CounterMetric, Integer>> Counters =
-            new ThreadLocal<ConcurrentMap<CounterMetric, Integer>>() {
-                @Override protected ConcurrentMap<CounterMetric, Integer> initialValue() {
-                    return metrics;
+    private static ConcurrentMetricCounters instance;
+    private static ConcurrentMap<CounterMetric, Integer> metrics = new ConcurrentHashMap<>();
```
zippolyte: Should this be a static attribute since it's not used in static methods ?

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
```diff
@@ -2,13 +2,31 @@
...
+    public static ReentrantLock lock = new ReentrantLock();
+
+    private ConcurrentMetricCounters(){}
+
+    public static ConcurrentMetricCounters getInstance(){
+        if(instance == null){
```
zippolyte: This top level if can probably be removed

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
```diff
@@ -2,13 +2,31 @@
...
+            synchronized (ConcurrentMetricCounters.class) {
```
zippolyte: Shouldn't we make the entire method `synchronized` ?

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
```diff
@@ -76,28 +90,51 @@ public boolean sendEvent(JSONObject payload) {
...
+                while(!ok) {
+                    logger.warning("Couldn't increment counter " + name + " with value " + (previousValue + 1) +
+                            " previousValue = " + previousValue);
+                    previousValue = counters.get(counterMetric);
+                    ok = counters.replace(counterMetric, previousValue, previousValue + 1);
+                }
             }
+            previousValue = previousValue == null ? 0 : previousValue;
+            logger.fine("Counter " + name + " updated from previousValue " + previousValue + " to "
+                    + (previousValue + 1));
```
zippolyte: Why not make this a synchronized method of the `ConcurrentMetricCounters` object like , and remove that manual lock ?

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
```diff
@@ -76,28 +90,51 @@ public boolean sendEvent(JSONObject payload) {
...
     public void flushCounters() {
-        ConcurrentMap<CounterMetric, Integer> counters = ConcurrentMetricCounters.Counters.get();
-        ConcurrentMetricCounters.Counters.set(new ConcurrentHashMap<CounterMetric, Integer>());
-
+        ConcurrentMetricCounters.lock.lock();
+        ConcurrentMap<CounterMetric, Integer> counters;
+        try {
+            ConcurrentMetricCounters countersInstance = ConcurrentMetricCounters.getInstance();
+            counters = countersInstance.get();
+            countersInstance.reset();
```
zippolyte: Same here, if you had a synchronized method like `getAndReset`, you could get rid of the manual lock

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
gzussa: Since it is a singleton it doesn't really matter I believe. It should enforce the fact that we should only have one instance per JVM.

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
gzussa: This is there for performance reasons. If this method is called often, `synchronized` can have some per implications. I don't expect this perf improvement to have huge consequences but still.

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
gzussa: see my comment above

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
gzussa: synchronized would prevent the same method to be called at the same time. Here with my lock I prevent we don't call both method at the same time

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
gzussa: see my comment above. I like the `getAndReset` method idea though :)

gzussa: Actually i won't do the `getAndReset` method. I initially thought I would have one in order to group the following logic: 
```
ConcurrentMetricCounters countersInstance = ConcurrentMetricCounters.getInstance();
counters = countersInstance.get();
```
Maybe i will create a Utilities method in another PR

src/main/java/org/datadog/jenkins/plugins/datadog/clients/DatadogHttpClient.java
zippolyte: Yeah that's my point, if you had a synchronized method that did the entire thing, you'd be fine, same idea as the `getAndReset`. 

So IMO, you shouldn't even be able to manually update the counters instance from outside of the class. 
The only way for you to obtain and manipulate the counters, should be when you are sure you will not do anymore operations on them, and so you would call `getAndReset` and do whatever you want with it at this point.

IOW, no `get` method, only a synchronized `increment` and a synchronized `getAndReset`. WDYT ? I believe it would make everything safer.

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
```diff
@@ -0,0 +1,72 @@
...
+        countersInstance.resetCounters();
+    }
+
+    private void resetCounters(){
+        counters = new ConcurrentHashMap<>();
+    }
+
+    private ConcurrentMap<CounterMetric, Integer> getCounters(){
+        return counters;
+    }
```
zippolyte: Are these really needed ?

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public static ConcurrentMap<CounterMetric, Integer> get(){
+        ConcurrentMetricCounters countersInstance = ConcurrentMetricCounters.getInstance();
+        return countersInstance.getCounters();
+    }
+
+    public static void reset(){
+        ConcurrentMetricCounters countersInstance = ConcurrentMetricCounters.getInstance();
+        countersInstance.resetCounters();
+    }
```
zippolyte: Can we remove or make these private as well

src/main/java/org/datadog/jenkins/plugins/datadog/clients/ConcurrentMetricCounters.java
gzussa: I prefer to use method rather than direct access. Not necessary but i prefer to keep it. We can remove later if it is really confusing.

Issue #3, opened 2015-08-31
Setup proper unit tests


JohnLZeller:

JohnLZeller:
Run through java [checkstyle](http://checkstyle.sourceforge.net/config.html) based on sun and google styles. Use their xmls to run checkstyle against.

PR #178, opened 2019-11-29
enable checks

gzussa:
Fixes https://github.com/DataDog/jenkins-datadog-plugin/issues/3

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,29 +97,4 @@
     </plugins>
   </build>
 
-  <reporting>
-    <plugins>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -97,29 +97,4 @@
     </plugins>
   </build>
 
-  <reporting>
```
zippolyte: I'm confused, why removing all that if we enable then in the pipeline ?

gzussa: The pipeline uses its own libs.

Issue #21, opened 2015-10-30
Add slaves statistics


awasilyev:
Start, stop event, initialization time...

JohnLZeller:
Thanks for the issue ticket @awasilyev ! I'm prioritizing this for one of the next releases.

PR #181, opened 2019-12-06
Adding more events

gzussa:
- Fixes https://github.com/DataDog/jenkins-datadog-plugin/issues/21
- See readme for metric, event and config addition/changes

Backward Compatibility Notes:
- Instead of sending `null` as default value for some tags we now send `unknown`
- Event messages got changed.


src/main/java/org/datadog/jenkins/plugins/datadog/listeners/DatadogItemListener.java
```diff
@@ -0,0 +1,131 @@
...
+    public void onUpdated(Item item) {
+        onCRUD(item, ItemCRUDEventImpl.UPDATED);
+    }
+
+    private void onCRUD(Item item, String action) {
+        try {
+            final boolean emitSystemEvents = DatadogUtilities.getDatadogGlobalDescriptor().isEmitSystemEvents();
+            if (!emitSystemEvents) {
+                return;
+            }
```
zippolyte: Is there a way to have this check in a top-level place and avoid having this once per Listener method ?
Like at instantiation for example, and we wouldn't even register the listeners at all ?

src/main/java/org/datadog/jenkins/plugins/datadog/listeners/DatadogSaveableListener.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+
+/**
+ * This class registers an {@link SaveableListener} to trigger events and calculate metrics:
+ * - When an saveable gets changed, the {@link #onChange(Saveable, XmlFile)} method will be invoked.
+ */
+@Extension
+public class DatadogSaveableListener  extends SaveableListener {
```
zippolyte: What's this annotation ? Why other Listeners don't have it ?

src/main/java/org/datadog/jenkins/plugins/datadog/util/TagsUtil.java
```diff
@@ -0,0 +1,43 @@
...
+        }
+        return dest;
+    }
+
+    public static JSONArray convertTagsToJSONArray(Map<String, Set<String>> tags){
+        JSONArray result = new JSONArray();
+        for (String name : tags.keySet()) {
+            Set<String> values = tags.get(name);
+            for (String value : values){
+                result.add(String.format("%s:%s", name, value));
```
zippolyte: Let's deal with the case where a tag is just a key and has no value

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogGlobalConfiguration.java
```diff
@@ -242,6 +246,48 @@ public void setWhitelist(final String jobs) {
...
+
+        for (String globalTagsLine : globalTagsLines) {
+            List<String> tagList = DatadogUtilities.cstrToList(globalTagsLine);
+            if (tagList.isEmpty()) {
+                continue;
+            }
+
+            for (int i = 1; i < tagList.size(); i++) {
+                String[] tagItem = tagList.get(i).split(":");
+                if(tagItem.length == 2) {
```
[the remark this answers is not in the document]
gzussa: The tag wouldn't be added. 
Indeed empty value tags are valid from a DD perspective but not supported here.

src/main/java/org/datadog/jenkins/plugins/datadog/events/BuildStartedEventImpl.java
```diff
@@ -23,29 +20,25 @@ public BuildStartedEventImpl(BuildData buildData, Map<String, Set<String>> build
...
-        String message = "%%% \n [Follow build #" +
-                number +
-                " progress](" +
-                builddata.getBuildUrl("unknown") +
-                ") " +
-                getFormattedDuration() +
-                " \n %%%";
+        // eg: [User <userId> started the [job <jobName> with build number #<buildNumber>] (1sec)"
+        String message = "%%% \n User " + userId + " started the [job " + jobName + " with build number #" +
+                buildNumber + "](" + buildUrl + ") " + getFormattedDuration() + " \n %%%";
```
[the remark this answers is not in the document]
gzussa: I am not sure either. In all cases, one more reason to bump a major with this.

src/main/java/org/datadog/jenkins/plugins/datadog/listeners/DatadogComputerListener.java
```diff
@@ -0,0 +1,189 @@
...
+            client.incrementCounter("jenkins.computer.online", hostname, tags);
+
+            logger.fine("End DatadogComputerListener#onOnline");
+        } catch (Exception e) {
+            logger.warning("Unexpected exception occurred - " + e.getMessage());
+        }
+    }
+
+    @Override
+    public void onOffline(@Nonnull Computer computer, @CheckForNull OfflineCause cause) {
```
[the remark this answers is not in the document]
gzussa: Same signature as the parent method

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogGlobalConfiguration.java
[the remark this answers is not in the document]
gzussa: Actually i will add this in another PR.

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -390,4 +374,66 @@ public static Boolean isValidHostname(String hostname) {
...
+            }
+            result.put("node_label", nodeLabelsValues);
+        }
+
+        return result;
+    }
+
+    public static String getNodeName(Computer computer){
+        if (computer instanceof Jenkins.MasterComputer) {
+            return "master";
```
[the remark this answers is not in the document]
gzussa: From what i understand a node is necessarily a non master computer in the Jenkins universe. so the `master` computer has no node name.

src/main/java/org/datadog/jenkins/plugins/datadog/listeners/DatadogSaveableListener.java
gzussa: Good catch i missed them. It registers the listener in Jenkins.

src/main/java/org/datadog/jenkins/plugins/datadog/listeners/DatadogItemListener.java
gzussa: Since i use `return`, i don't think so. I could create an exception but i will still have to handle it when it would be thrown.

src/main/java/org/datadog/jenkins/plugins/datadog/util/TagsUtil.java
gzussa: Will add this as part of another PR. See my previous comment

README.md
```diff
@@ -8,70 +8,147 @@ There is a [Jenkins-CI Wiki page](https://wiki.jenkins-ci.org/display/JENKINS/Da
...
+  - Associated rate metric: `jenkins.user.access_denied`
+* User loggedOut
+  - Triggered on `SecurityListener#loggedOut`
+  - Associated rate metric: `jenkins.user.logout`
+
+### Metrics
+
+| Metric Name                          | Description                                                   | Default Tags                               |
+|--------------------------------------|---------------------------------------------------------------|--------------------------------------------|
+| jenkins.computer.launch_failure      | Rate of computer launch failures                              |                                            |
```
l0k0ms:
```suggestion
| `jenkins.computer.launch_failure`      | Rate of computer launch failures.                              |                                            |
```

README.md
```diff
@@ -8,70 +8,147 @@ There is a [Jenkins-CI Wiki page](https://wiki.jenkins-ci.org/display/JENKINS/Da
...
+| jenkins.computer.offline             | Rate of computer going offline                                |                                            |
```
l0k0ms:
```suggestion
| `jenkins.computer.offline`             | Rate of computer going offline.                                |                                            |
```

Issue #110, opened 2018-01-10
Append the default parameter "Add tags from file in workspace" in general plugin configuration


cazorla19:
To be clear I'll explain the issue first.

Let's assume that we have a Github organization/Bitbucket organization folder where all Jenkins jobs are created automatically for every new repository with Jenkinsfile or repo branches. This setup is pretty popular for now. But how can we tag the builds for Datadog? For now there is no clear option for this. Every job has Datadog tagging parameters, but they couldn't be modified because the organization folder jobs are immutable. Also the tags couldn't be appended at any folder level.

Also here is the issue out of such setup. Even if we manually create Jenkins job, we have to always make sure that the "Add tags from file in workspace" option is filled and we can't inherit any default settings. Which sometimes seems pretty useful due to organisation standards (e.g. have a `tags.dd` file in every repo/project with tags declaration). And, if you want to migrate to Datadog - it means that you have to setup this parameter manually on all existing jobs (that's probably wrong, but all my attempts to do this through XML reconfig&reload and Groovy scripts were failed).

If we could append the option in "Manage Jenkins &rarr; Configure System &rarr; Datadog Plugin" the option to declare the default file in workspace and inherit this parameter for every job. This feature would make the tagging setup much easier. 

Please, let me know if something isn't clear at this issue.

sjenriquez:
@cazorla19 Sorry for the slow reply. Your idea for adding a default/global workspace file path makes sense. We'll try to add this feature as soon as possible, meanwhile if you happen to work on this please submit a PR, thanks!

PR #182, opened 2019-12-12
single tag global conf

gzussa:
- Support tags with no values
- Fixes https://github.com/DataDog/jenkins-datadog-plugin/issues/110


src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -143,6 +147,13 @@ private static String getOS() {
                         Set<String> tagValues = tags.containsKey(tagName) ? tags.get(tagName) : new HashSet<String>();
                         tagValues.add(tagValue.toLowerCase());
                         tags.put(tagName, tagValues);
+                    } else if(tagItem.length == 1) {
```
zippolyte: we should also consider the fact that there could me more than 1 colon in the tags. To do so, let's simply do `split(":", 2)`

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -168,14 +179,21 @@ private static String getOS() {
                 continue;
             }
 
-            for (int i = 0; i < tagList.size(); i++) {
+            for(int i = 0; i < tagList.size(); i++) {
                 String[] tagItem = tagList.get(i).replaceAll(" ", "").split(":");
```
zippolyte:
```suggestion
                String[] tagItem = tagList.get(i).replaceAll(" ", "").split(":", 2);
```

src/main/java/org/datadog/jenkins/plugins/datadog/DatadogUtilities.java
```diff
@@ -271,14 +289,19 @@ private static boolean isJobWhitelisted(final String jobName) {
             }
             for (int i = 0; i < tagList.size(); i++) {
                 String tag = tagList.get(i).replaceAll(" ", "");
-                String[]expanded = envVars.expand(tag).split("=");
-                if (expanded.length > 1) {
+                String[] expanded = envVars.expand(tag).split("=");
```
zippolyte: Should we do the same here  ? I'm confused about what this function actually does

PR #183, opened 2019-12-13
Add dogStatsDClient

gzussa:
- Fixes https://github.com/DataDog/jenkins-datadog-plugin/issues/34

Related to https://github.com/DataDog/jenkins-datadog-plugin/pull/156 and https://github.com/DataDog/jenkins-datadog-plugin/pull/132


src/main/java/org/datadog/jenkins/plugins/datadog/clients/DogStatsDClient.java
```diff
@@ -0,0 +1,210 @@
...
+            synchronized (DatadogHttpClient.class) {
+                if(instance == null){
+                    instance = new DogStatsDClient( url, port);
+                }
+            }
+        }
+
+        // We reset param just in case we change values
+        instance.setApiKey(null);
+        instance.setUrl(url);
```
zippolyte: I don't understand this

gzussa: You don't understand because it is indeed wrong 😆

